PR #72, opened 2022-04-05
Set up lerna and add test infrastructure

### LINKED ISSUE #2: Setup Lerna Repo

### LINKED ISSUE #65: [ Tracking] Set up tests infrastructure
- [ ] Add node and browser tests to abstractions
- [ ] Add node and browser tests to fetch
- [ ] Add node and browser tests to serialization json
- [ ] Add node and browser tests to serialization text
- [ ] Add node and browser tests to authorization azure
- [ ] Add integrated run time tests

nikithauc:
fixes #2 
fixes #65


package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "root",
```
baywet, line 2: shouldn't we add a dependabot entry for this new package file?

.github/workflows/build_test_validate.yml
```diff
@@ -0,0 +1,30 @@
...
+        steps:
+        - uses: actions/checkout@v3
+        - name: Use Node.js ${{ matrix.node-version }}
+          uses: actions/setup-node@v3
+          with:
+            node-version: ${{ matrix.node-version }}
+        - run: yarn install --frozen-lockfile
+        - run: yarn run build
+        - run: lerna run test:integrated
+        - run: lerna run test --parallel
```
baywet, line 26: isn't this missing a deploy job?

.github/workflows/build_test_validate.yml
```diff
@@ -0,0 +1,30 @@
...
+        strategy:
+          matrix:
+            node-version: [14.x, 16.x]
+        steps:
+        - uses: actions/checkout@v3
+        - name: Use Node.js ${{ matrix.node-version }}
+          uses: actions/setup-node@v3
+          with:
+            node-version: ${{ matrix.node-version }}
+        - run: yarn install --frozen-lockfile
```
baywet, line 23: I'm curious to why this has been switched to yarn? is lerna pushing that requirement on us?

packages/abstractions/package.json
```diff
@@ -2,21 +2,25 @@
     "name": "@microsoft/kiota-abstractions",
     "version": "1.0.0-preview.1",
```
baywet, line 3: my guess is the corresponding package.lock has been deleted for two reasons:
- switch to yarn
- lerna has a single lock at the root
But I just wanted to confirm?

packages/abstractions/rollup.config.js
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.
+ * See License in the project root for license information.
+ * -------------------------------------------------------------------------------------------
+ */
+
+import resolve from "@rollup/plugin-node-resolve";
+import commonjs from "@rollup/plugin-commonjs";
+
+const config = [{
+    input: ["dist/es/test/browser/index.js"],
```
baywet, line 12: is this rollup configuration only used for tests? (just want to validate the paths)

packages/serialization/json/package.json
```diff
@@ -31,17 +35,6 @@
...
-        "@typescript-eslint/parser": "^5.16.0",
-        "eslint": "^8.11.0",
-        "eslint-config-prettier": "^8.5.0",
-        "eslint-plugin-prettier": "^4.0.0",
-        "eslint-plugin-simple-import-sort": "^7.0.0",
-        "prettier": "^2.6.0",
-        "typescript": "^4.6.2"
-    },
     "dependencies": {
         "@microsoft/kiota-abstractions": "^1.0.0-preview.1",
```
baywet, line 39: shouldn't this dep be tokenized by lerna?

baywet (commented):
thanks for starting this. a couple of comments/questions

.github/workflows/build_test_validate.yml
nikithauc: Versioning and deploy job will be a separate workflow.

packages/abstractions/package.json
nikithauc: Yeah right. We do a single yarn install at the root where the locked versions are maintained

packages/abstractions/rollup.config.js
nikithauc: Yeah for now. 

The path here is the index for the test files. The src and tests are rolled up and tested with karma

packages/serialization/json/package.json
nikithauc: lerna versioning takes care of it. 

The command lerna version updates the dependencies

package.json
nikithauc: Yeah right

.github/workflows/build_test_validate.yml
nikithauc: yarn workspaces is optimal for dependency management and linking local dependencies.  It is popularly coupled with lerna for multi package management.

.github/workflows/build_test_validate.yml
baywet: what are you planning to be the trigger of releases? a tag creation? push on main? something else?

.github/workflows/build_test_validate.yml
baywet: I haven't used yarn in ages (back when it used to be much faster than npm). We should probably update the readme accordingly so people know which tools they need to get started.

package.json
```diff
@@ -0,0 +1,47 @@
...
+  "workspaces": [
+    "packages/test/",
+    "packages/abstractions",
+    "packages/http/*",
+    "packages/serialization/json/",
+    "packages/serialization/text/"
+  ],
+  "scripts": {
+      "build": "npm run build:cjs && npm run build:es",
+      "build:cjs": "tsc -b packages/test/tsconfig.cjs.json",
```
baywet, line 43: also it seems that the default build scripts that'll be used for publishing relies on the test tsconfig file. this probably needs to be updated.

.github/workflows/build_test_validate.yml
nikithauc: The plan is to use lerna version which updates the package version as well as updates the dependencies and creates the tags. 
This way the dependent packages are in sync. If the package versions are not in sync, then a dependent packages looks for the parent package in npm instead of using the local code. With manual version updates, it is easy to miss updating version numbers in every place. 

The versioning part will be automated in #93. 

The process is to run:
`lerna version`  plus a customized automated process followed by
`lerna publish` publishes to npm.

package.json
nikithauc: The build scripts build all the packages as they are linked to the test tsconfig file using project references. This ensures the the code is building with the latest changes and in the right order, that is, example abstractions is built before http.

.github/workflows/build_test_validate.yml
baywet: so if I understand correctly, between the time we merge this PR and the time we merge the PR for #93, we won't be able to release?

package.json
nikithauc: Done

.github/workflows/build_test_validate.yml
nikithauc: We will be able to release. I will manually update the versions using lerna commands and publish off those tags. #93 includes some work to update version with custom suffixes. 

The main objective of this PR is too have the initial set up and enable easy testing and validation to include further PRs. 

publish and versioning wont be blocked with these changes.

baywet: I might be missing something, but once you push the new tags, what's going to perform the npm publish?

baywet: resolving as this was just added, thanks!

baywet (approved):

baywet (approved):

PR #87, opened 2022-04-08
adds support for decoding query parameters names

baywet:
related https://github.com/microsoft/kiota/issues/1445


packages/http/fetch/src/middlewares/browser/middlewareFactory.ts
```diff
@@ -29,12 +29,7 @@ export class MiddlewareFactory {
 	public static getDefaultMiddlewareChain(customFetch: (request: string, init: RequestInit) => Promise<Response> = fetch as any): Middleware[] {
 		// Browsers handles redirection automatically and do not require the redirectionHandler
 
-		const middlewareArray: Middleware[] = [];
-		const retryHandler = new RetryHandler(new RetryHandlerOptions());
-		middlewareArray.push(retryHandler);
-
-		middlewareArray.push(new CustomFetchHandler(customFetch));
-
+		const middlewareArray: Middleware[] = [new RetryHandler(), new ParametersNameDecodingHandler(), new CustomFetchHandler(customFetch)];
```
nikithauc, line 32: The problem with using a handler is that it can be easily  missed in the custom middleware chains.

nikithauc, line 32: The problem with using a handler is that it can be easily  missed in the custom middleware chains.

packages/abstractions/src/requestInformation.ts
```diff
@@ -131,7 +131,14 @@ export class RequestInformation {
    */
   public setQueryStringParametersFromRawObject = (q: object): void => {
     Object.entries(q).forEach(([k, v]) => {
-      this.queryParameters[k] = v;
+      let key = k;
+      if ((q as any).getQueryParameter) {
+        const serializationKey = (q as any).getQueryParameter(key) as string;
```
nikithauc, line 136: Please add  an integrated test  in packages/test which passes query params and verify if the real time tests work with query params

packages/http/fetch/src/middlewares/browser/middlewareFactory.ts
baywet: Are you saying we should not use a handler to decode the parameters names?

packages/http/fetch/src/middlewares/parametersNameDecodingHandler.ts
```diff
@@ -0,0 +1,53 @@
...
+	 * @param {Record<string, RequestOption>} [requestOptions] - The request options
+	 * @returns A Promise that resolves to nothing
+	 */
+	public execute(url: string, requestInit: RequestInit, requestOptions?: Record<string, RequestOption>): Promise<Response> {
+		let currentOptions = this.options;
+		if (requestOptions && requestOptions[ParametersNameDecodingHandlerOptionsKey]) {
+			currentOptions = requestOptions[ParametersNameDecodingHandlerOptionsKey] as ParametersNameDecodingHandlerOptions;
+		}
+		let updatedUrl = url;
+		if (currentOptions && currentOptions.enable && url.indexOf("%") > -1 && currentOptions.charactersToDecode && currentOptions.charactersToDecode.length > 0) {
```
nikithauc, line 46: I am familiar with some cases where the Graph API expects the some characters to be double encoded. 

Following is an example: 

https://github.com/microsoftgraph/microsoft-graph-toolkit/pull/1197/files#diff-7d9ac68ab30b3e76b018d22a3f46aa743444fce208ce9f461dcc8b009e275aa9R231

Would this avoid decoding or manipulating the double encoded values which are expected by Graph in some cases?

packages/abstractions/src/requestInformation.ts
nikithauc: Add a test  to verify https://graph.microsoft.com/v1.0/users?$search=%22displayName:%2523%22&$count=true

nikithauc:
Please  rebase to take the test set up updates

packages/http/fetch/test/common/middleware/parametersNameDecodingHandler.ts
```diff
@@ -0,0 +1,53 @@
...
+	data.forEach((entry) => {
+		it(`Should decode the parameters names ${entry.expected}`, async () => {
+			const fetchRequestInit = {
+				method: "GET",
+			};
+			const requestInformationOptions = {};
+			handler.next = new CallBackMiddleware((url: string) => {
+				assert.equal(url, entry.expected);
+			});
+			await handler["execute"](entry.input, fetchRequestInit, requestInformationOptions);
```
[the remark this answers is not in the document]
baywet: done. Thanks for the callout.

packages/http/fetch/src/middlewares/browser/middlewareFactory.ts
baywet: resolving this comment as it it a duplicate of the previous one.

packages/http/fetch/src/middlewares/parametersNameDecodingHandler.ts
baywet: yes, it might decode values as implemented today. However we're only decoding `~ . - $`, and the caller can always disable the middleware, or change the set of characters, either all together or on a per request base.
We've had a long discussion about these cases with @darrelmiller before deciding on an implementation. Effectively the service should never expect a double encoded value. single encoding is fine though.

nikithauc (approved):

packages/http/fetch/src/middlewares/parametersNameDecodingHandler.ts
nikithauc: Can you please add a document in this repo something like encoding-decoding.md mentioning this features 
and also create an issue in Graph JS repos to test this feature and mark the issue as test.

baywet: done https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/753

PR #202, opened 2022-05-23
Update test generated code  with model interfaces

nikithauc:
Generation based on https://github.com/microsoft/kiota/pull/1559


packages/test/generatedCode/models/microsoft/graph/attachmentImpl.ts
```diff
@@ -0,0 +1,64 @@
...
+    /** The attachment's file name. */
+    public name?: string | undefined;
+    /** The length of the attachment in bytes. */
+    public size?: number | undefined;
+    /**
+     * Instantiates a new attachment and sets the default values.
+     * @param attachmentParameterValue 
+     */
+    public constructor(attachmentParameterValue?: Attachment | undefined) {
+        super();
```
baywet, line 21: shouldn't the parameter be passed to the parent constructor as well to get the properties from the parent assigned as well?

packages/test/generatedCode/models/microsoft/graph/attachment.ts
```diff
@@ -1,117 +1,14 @@
...
-    /**
-     * Sets the size property value. The length of the attachment in bytes.
-     * @param value Value to set for the size property.
-     */
-    public set size(value: number | undefined) {
-        this._size = value;
-    };
+export interface Attachment extends Entity{
+    /** The MIME type. */
+    contentType?:string | undefined;
```
baywet, line 5: nit: lack of space between the property name and the type

packages/test/generatedCode/models/microsoft/graph/attachmentImpl.ts
```diff
@@ -0,0 +1,64 @@
...
+    };
+    /**
+     * Serializes information the current object
+     * @param writer Serialization writer to use to serialize this model
+     */
+    public serialize(writer: SerializationWriter) : void {
+        if(!writer) throw new Error("writer cannot be undefined");
+        super.serialize(writer);
+        if(this.contentType){
+        writer.writeStringValue("contentType", this.contentType);
```
baywet, line 49: indent

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZone.ts
```diff
@@ -1,78 +1,9 @@
...
-    /**
-     * Sets the timeZone property value. Represents a time zone, for example, 'Pacific Standard Time'. See below for more possible values.
-     * @param value Value to set for the timeZone property.
-     */
-    public set timeZone(value: string | undefined) {
-        this._timeZone = value;
-    };
+export interface DateTimeTimeZone{
+    /** Stores additional data not described in the OpenAPI description found when deserializing. Can be used for serialization as well. */
+    additionalData?:Record<string, unknown>;
```
baywet, line 4: this property doesn't need to be re-defined

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZone.ts
```diff
@@ -1,78 +1,9 @@
...
-        return this._timeZone;
-    };
-    /**
-     * Sets the timeZone property value. Represents a time zone, for example, 'Pacific Standard Time'. See below for more possible values.
-     * @param value Value to set for the timeZone property.
-     */
-    public set timeZone(value: string | undefined) {
-        this._timeZone = value;
-    };
+export interface DateTimeTimeZone{
```
baywet, line 2: the interface should inherit from IParsable and IAdditionalPropertyHolder

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZoneImpl.ts
```diff
@@ -0,0 +1,44 @@
+import {DateTimeTimeZone} from './dateTimeTimeZone';
+import {AdditionalDataHolder, Parsable, ParseNode, SerializationWriter} from '@microsoft/kiota-abstractions';
+
+export class DateTimeTimeZoneImpl implements AdditionalDataHolder, DateTimeTimeZone, Parsable {
```
baywet, line 4: assuming the model interface holds the inheritance to Parsable and AdditionalPropertyHolder, this doesn't need to be repeated here

baywet (changes requested):

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZone.ts
nikithauc: IParsable and IAdditionalPropertyHolder has required parameters which should be optional in the the model interface. 

In that case, an example as follows throws the errors that `additionalData` or `getFieldDeserializer` is not mentioned. 
```
const message = {
 subject: ""
}
```

The current implementation avoids that. 

The other option is to use `implements  Partial<IParsable> and Partial<AddnHolder>.

nikithauc: I think the second option works well. What do you think?

baywet: Yeah that's a good suggestion. An alternative would be to have the request body parameters and the return types to be intersection types: ModelInterface & Parsable & AdditionalPropertyHolder, so people know they also get those extra properties and can provide additional properties.

packages/test/generatedCode/models/microsoft/graph/attachmentImpl.ts
baywet: I can see that you updated to the following.

```typescript
super({
            additionalData: extensionParameterValue?.additionalData,
            id: extensionParameterValue?.id,
        });
```

Why not do that instead 

```
super(extensionParameterValue);
```

Since the derived type will fulfill the base type interface? (to avoid generating too much code)

nikithauc:
This PR is up-to-date reflecting the generation. Marking this as draft to merge it after https://github.com/microsoft/kiota/pull/1559 is merged.

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZone.ts
nikithauc: Resolved

packages/test/generatedCode/models/microsoft/graph/dateTimeTimeZone.ts
nikithauc: This is happening in the current main generation and not introduced in this PR. Marking it as a separate issue.

nikithauc: https://github.com/microsoftgraph/msgraph-sdk-typescript/blob/4dfd1260694a59f2e09cb60fb6a8ee99990bcf2d/src/models/dateTimeTimeZone.ts#L5

PR #407, opened 2022-10-27
Added Kiota AuthProvider for SPFx solutions

luismanez:
This PR adds a new Kiota AuthProvider implementation to be used in SPFx solutions. The provider re-uses the existing Token provider provided by the same SharePoint Framework infrastructure. This provider can also be used for generated MS Graph Kiota clients.


packages/authentication/spfx/README.md
```diff
@@ -0,0 +1,62 @@
+[![npm version badge](https://img.shields.io/npm/v/@microsoft/kiota-authentication-spfx?color=blue)](https://www.npmjs.com/package/@microsoft/kiota-authentication-spfx)
+
+The Kiota Authentication SPFx Library is an implementation to authenticate HTTP requests in SPFx solutions. This Authentication library, re uses some infrastructure from the SPFx context to obtain a valid token for the required API.
+
+It also works for consuming MS Graph API, as Graph is just another Azure AD protected API.
```
baywet, line 5:
```suggestion
It also works for consuming Microsoft Graph, as it is just another Azure Active Directory protected API.
```

packages/authentication/spfx/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+        myCustomApiClient.teams.get().then(teams => {
+            // deal with returned data
+        })
+        .catch(e => {console.log(e)});
+      })
+      .catch(e => {console.log(e)});
+```
+
+If you have generated a Kiota client for MS Graph API, you can also use this Auth provider in a similar way:
```
baywet, line 41:
```suggestion
If you have generated a Kiota client for Microsoft Graph, you can also use this Auth provider in a similar way:
```

packages/authentication/spfx/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```ts
+    this.props.aadTokenProviderFactory.getTokenProvider()
+      .then((tokenProvider: AadTokenProvider): void => {
+
+        const authProvider =
+          new AzureAdSpfxAuthenticationProvider(
+            tokenProvider, 
+            "https://graph.microsoft.com");
```
baywet, line 50: shouldn't this be just the hostname, or the .default permission scope?

packages/authentication/spfx/package-lock.json
```diff
@@ -0,0 +1,462 @@
+{
+    "name": "@microsoft/kiota-authentication-spfx",
```
baywet, line 2: we're using lerna here, which means the libs don't have a package lock (managed by lerna).
remove this file.
add any additional package this lib needs to the main package.json with yarn

packages/authentication/spfx/package.json
```diff
@@ -0,0 +1,47 @@
...
+        "OpenApi",
+        "Azure",
+        "Authentication",
+        "OAuth",
+        "SPFx",
+        "AzureAD",
+        "Graph",
+        "SharePoint"
+    ],
+    "author": "Luis Mañez",
```
baywet, line 35: please use the same author as the other packages and add yourself as a contributor instead.

packages/authentication/spfx/tsconfig.base.json
```diff
@@ -0,0 +1,23 @@
...
+        "declaration": true,
+        "declarationMap": true,
+        "composite": true,
+        "paths": {
+            "@microsoft/kiota-abstractions": [
+                "../../abstractions"
+            ],
+        },
+
+        "skipLibCheck": true, // Otherwise, raises different issues with SPFx dependencies types (i.e: cannot find IInternalAadTokenProvider)
```
baywet, line 21:
```suggestion
        "skipLibCheck": true
```
should be fine from the documentation, thanks for pointing out

packages/authentication/spfx/README.md
luismanez: yup, my bad! if you pass it, should be just the hostname, but by default is set to Microsoft Graph host, so in this case is not required.

packages/authentication/spfx/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "contributors": [
+        {"name":"Luis Mañez","url":"https://github.com/luismanez"}
+    ],
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/microsoft/kiota-typescript/issues"
+    },
+    "homepage": "https://github.com/microsoft/kiota-typescript#readme",
+    "dependencies": {
+        "@azure/core-auth": "^1.3.2",
```
baywet, line 45:
```suggestion
```
I don't think this one it needed for this package

baywet (changes requested):

packages/authentication/spfx/package.json
luismanez: yeah, agreed, I think was there when I was testing the best way to mock the SPFx aadTokenProvider. thanks!

baywet (dismissed):
Thanks for the great contribution and for addressing the feedback so quickly.

If you're looking to contribute more, a teams fx auth provider would be a great follow up to this and should not require any additional learning from you on the kiota side (low hanging fruit)

baywet (dismissed):

baywet (dismissed):

baywet (approved):

sebastienlevert:
You can get some inspiration for the existing auth provider in MGT here : https://github.com/microsoftgraph/microsoft-graph-toolkit/tree/main/packages/providers/mgt-teamsfx-provider

This for this awesome contribution!

luismanez:
Thanks for your help both. Will take a look to that teams fx thing and see what I can do :)

baywet:
The package is up! https://www.npmjs.com/package/@microsoft/kiota-authentication-spfx
💪

PR #427, opened 2022-11-11
feature/observability

baywet:
adds tracing support through open telemetry
spec https://github.com/microsoftgraph/msgraph-sdk-design/blob/dm/logging/Observability.md
related https://github.com/microsoft/kiota/issues/618

Had to be put on pause due to other priorities, but I finally got to wrap it up :)

### DIFF
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -1,3 +1,4 @@
+import { trace } from "@opentelemetry/api";
 import * as urlTpl from "uri-template-lite";
 
 import { DateOnly } from "./dateOnly";
@@ -94,8 +95,10 @@ export class RequestInformation {
       delete this._requestOptions[option.getKey()];
     });
   }
-  private static binaryContentType = "application/octet-stream";
-  private static contentTypeHeader = "Content-Type";
+  private static readonly binaryContentType = "application/octet-stream";
+  private static readonly contentTypeHeader = "Content-Type";
+  private static readonly tracerKey = "@microsoft/kiota-abstractions";
+  private static readonly requestTypeKey = "com.microsoft.kiota.request.type";
   /**
    * Sets the request body from a model with the specified content type.
    * @param value the models.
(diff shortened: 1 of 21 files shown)


baywet:
I recommend reviewers to check the "hide whitespace changes" option.
This is what a trace looks like in zipkin

image


if you want to try it out (in a node app)

```TypeScript
diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
      const resource =
      Resource.default().merge(
        new Resource({
          [SemanticResourceAttributes.SERVICE_NAME]: "service-name-here",
          [SemanticResourceAttributes.SERVICE_VERSION]: "0.1.0",
        })
      );

    const provider = new NodeTracerProvider({
        resource: resource,
    });
    const exporter = new ZipkinExporter({
    });
    provider.addSpanProcessor(new SimpleSpanProcessor(exporter));
    provider.addSpanProcessor(new SimpleSpanProcessor(new ConsoleSpanExporter()));
    provider.register();

    await trace.getTracer(
      'service-name-here'
    ).startActiveSpan('my-span', async (span) => {
       // do things with the client here
      span.end();
 });
```

packages/http/fetch/src/observabilityOptions.ts
```diff
@@ -0,0 +1,37 @@
+import { RequestOption } from "@microsoft/kiota-abstractions";
+
+export interface ObservabilityOptions {
+	includeEUIIAttributes: boolean;
+}
+export interface ObservabilityOptionsInternal {
```
nikithauc, line 6: Add summary for the `ObservabilityOptionsInternal`  and `ObservabilityOptions` especially stating the distinction

packages/abstractions/src/requestInformation.ts
```diff
@@ -108,20 +111,30 @@ export class RequestInformation {
...
-    }
-    if (Array.isArray(value)) {
-      writer.writeCollectionOfObjectValues(undefined, value);
-    } else {
-      writer.writeObjectValue(undefined, value);
-    }
-    this.setContentAndContentType(writer, contentType);
+    trace
+      .getTracer(RequestInformation.tracerKey)
+      .startActiveSpan("setContentFromParsable", (span) => {
```
nikithauc, line 116: Why are the code additions not optional? What if the library user does not require telemetry collection and opts out?

baywet: The general design guidance for libraries with open telemetry is to implement the tracing. If the library user sets collectors and exporters, it'll go where asked to go. If they don't, all of that just pipes to null. This way libraries don't have to implement settings which could be hard to control with nested of libs about enabling or not tracing.

packages/http/fetch/src/observabilityOptions.ts
baywet: done

nikithauc (approved):

packages/http/fetch/src/observabilityOptions.ts
```diff
@@ -0,0 +1,39 @@
+import { RequestOption } from "@microsoft/kiota-abstractions";
+
+/** Holds the tracing, metrics and logging configuration for the request adapter */
+export interface ObservabilityOptions {
+	includeEUIIAttributes: boolean;
+}
+/** Internal interface not meant to be used externally and designed to facilitate JSON usage of the other interface. */
+export interface ObservabilityOptionsInternal {
```
nikithauc, line 8: Remove the export here if not meant to be used outside this file scope

baywet: done

gavinbarron (approved):

PR #469, opened 2022-12-15
feature/form encoded

baywet:
related https://github.com/microsoft/kiota/issues/219

### DIFF
--- a/packages/serialization/form/src/formParseNode.ts
+++ b/packages/serialization/form/src/formParseNode.ts
@@ -31,7 +31,8 @@ export class FormParseNode implements ParseNode {
         }
       });
   }
-  private normalizeKey = (key: string): string => key.trim();
+  private normalizeKey = (key: string): string =>
+    decodeURIComponent(key).trim();
   public onBeforeAssignFieldValues: ((value: Parsable) => void) | undefined;
   public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
   public getStringValue = (): string => decodeURIComponent(this._rawString);
--- a/packages/serialization/form/src/formParseNode.ts
+++ b/packages/serialization/form/src/formParseNode.ts
@@ -1,0 +1,109 @@
+import {
+  AdditionalDataHolder,
+  DateOnly,
+  Duration,
+  Parsable,
+  ParsableFactory,
+  ParseNode,
+  TimeOnly,
+  toFirstCharacterUpper,
+} from "@microsoft/kiota-abstractions";
+
+export class FormParseNode implements ParseNode {
+  private readonly _fields: Record<string, string> = {};
+  /**
+   *
+   */
+  constructor(private readonly _rawString: string) {
+    if (!_rawString) {
+      throw new Error("rawString cannot be undefined");
+    }
+    _rawString
+      .split("&")
+      .map((x) => x.split("="))
+      .filter((x) => x.length === 2)
+      .forEach((x) => {
+        const key = this.normalizeKey(x[0]);
+        if (this._fields[key]) {
+          this._fields[key] += "," + x[1];
+        } else {
+          this._fields[key] = x[1];
+        }
+      });
+  }
+  private normalizeKey = (key: string): string => key.trim();
+  public onBeforeAssignFieldValues: ((value: Parsable) => void) | undefined;
+  public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
+  public getStringValue = (): string => decodeURIComponent(this._rawString);
+  public getChildNode = (identifier: string): ParseNode | undefined => {
+    if (this._fields[identifier]) {
+      return new FormParseNode(this._fields[identifier]);
+    }
+    return undefined;
+  };
+  public getBooleanValue = (): boolean =>
+    JSON.parse(this.getStringValue()) as boolean;
+  public getNumberValue = (): number => parseFloat(this.getStringValue());
+  public getGuidValue = (): string => this.getStringValue();
+  public getDateValue = (): Date => new Date(Date.parse(this.getStringValue()));
+  public getDateOnlyValue = () => DateOnly.parse(this.getStringValue());
+  public getTimeOnlyValue = () => TimeOnly.parse(this.getStringValue());
+  public getDurationValue = () => Duration.parse(this.getStringValue());
+  public getCollectionOfPrimitiveValues = <T>(): T[] | undefined => {
+    throw new Error(
+      `serialization of collections is not supported with URI encoding`
+    );
+  };
+  public getCollectionOfObjectValues = <T extends Parsable>(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _type: ParsableFactory<T>
+  ): T[] | undefined => {
+    throw new Error(
+      `serialization of collections is not supported with URI encoding`
+    );
+  };
+  public getObjectValue = <T extends Parsable>(type: ParsableFactory<T>): T => {
+    const result = type(this);
+    if (this.onBeforeAssignFieldValues) {
+      this.onBeforeAssignFieldValues(result);
+    }
+    this.assignFieldValues(result);
+    if (this.onAfterAssignFieldValues) {
+      this.onAfterAssignFieldValues(result);
+    }
+    return result;
+  };
+  public getEnumValues = <T>(type: any): T[] => {
+    const rawValues = this.getStringValue();
+    if (!rawValues) {
+      return [];
+    }
+    return rawValues.split(",").map((x) => type[toFirstCharacterUpper(x)] as T);
+  };
+  public getEnumValue = <T>(type: any): T | undefined => {
+    const values = this.getEnumValues(type);
+    if (values.length > 0) {
+      return values[0] as T;
+    } else {
+      return undefined;
+    }
+  };
+  private assignFieldValues = <T extends Parsable>(item: T): void => {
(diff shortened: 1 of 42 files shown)


packages/serialization/form/src/formParseNode.ts
```diff
@@ -0,0 +1,109 @@
...
+  public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
+  public getStringValue = (): string => decodeURIComponent(this._rawString);
+  public getChildNode = (identifier: string): ParseNode | undefined => {
+    if (this._fields[identifier]) {
+      return new FormParseNode(this._fields[identifier]);
+    }
+    return undefined;
+  };
+  public getBooleanValue = (): boolean =>
+    JSON.parse(this.getStringValue()) as boolean;
```
gavinbarron, line 45: Won't this throw an exception when it tries to JSON.parse the _rawString that was passed into the constructor?
Or is the intent that this would only ever be called on a childNode? Although the same issue still arises when the childNode isn't a boolean.

What should be the expected behavior if the parse throws?

baywet: The fact that there doesn't seem to be a try parse for boolean in JavaScript (true|false|undefined) makes me sad...
I had a look at what I did for the text provider, which was better but partial. So I fixed both implementations. Thanks for calling this out.

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+export class FormSerializationWriter implements SerializationWriter {
+  private readonly writer: string[] = [];
+  private static propertySeparator = `&`;
+  private depth = -1;
+  public onBeforeObjectSerialization: ((value: Parsable) => void) | undefined;
+  public onAfterObjectSerialization: ((value: Parsable) => void) | undefined;
+  public onStartObjectSerialization:
+    | ((value: Parsable, writer: SerializationWriter) => void)
+    | undefined;
+  public writeStringValue = (key?: string, value?: string): void => {
```
gavinbarron, line 19: If this is called with a falsy key and a truthy value then the value is written without a key or separator, is that valid?

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  public writeBooleanValue = (key?: string, value?: boolean): void => {
+    value !== null &&
+      value !== undefined &&
+      this.writeStringValue(key, `${value}`);
+  };
+  public writeNumberValue = (key?: string, value?: number): void => {
+    value && this.writeStringValue(key, `${value}`);
+  };
+  public writeGuidValue = (key?: string, value?: string): void => {
+    value && this.writeStringValue(key, `${value}`);
```
gavinbarron, line 36: why are we adding the cost of interpolating values that are already strings in here?

packages/serialization/form/test/browser/index.ts
```diff
@@ -0,0 +1 @@
+export * from "../common/formParseNodeFactory";
```
gavinbarron, line 1: What purpose is this file serving?

baywet: runs the same test but in a browser context I believe, replicated from the JSON serialization provider. maybe @nikithauc has more context here.

packages/serialization/form/src/formSerializationWriter.ts
baywet:
```suggestion
    value && this.writeStringValue(key, value);
```

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  };
+  public writeDateValue = (key?: string, value?: Date): void => {
+    value && this.writeStringValue(key, `${value.toISOString()}`);
```
baywet, line 39:
```suggestion
    value && this.writeStringValue(key, value.toISOString());
```

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  };
+  public writeDateOnlyValue = (key?: string, value?: DateOnly): void => {
+    value && this.writeStringValue(key, `${value.toString()}`);
```
baywet, line 42:
```suggestion
    value && this.writeStringValue(key, value.toString());
```

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  };
+  public writeTimeOnlyValue = (key?: string, value?: TimeOnly): void => {
+    value && this.writeStringValue(key, `${value.toString()}`);
```
baywet, line 45:
```suggestion
    value && this.writeStringValue(key, value.toString());
```

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  };
+  public writeDurationValue = (key?: string, value?: Duration): void => {
+    value && this.writeStringValue(key, `${value.toString()}`);
```
baywet, line 48:
```suggestion
    value && this.writeStringValue(key, value.toString());
```

packages/serialization/form/test/common/formParseNode.ts
```diff
@@ -0,0 +1,52 @@
...
+    "userPrincipalName=MeganB@M365x214355.onmicrosoft.com&" +
+    "birthday=2017-09-04&" +
+    "id=48d31887-5fad-4d73-a9f5-3c356e68a038";
+  it("getsEntityValueFromForm", () => {
+    const parseNode = new FormParseNode(testUserForm);
+    const testEntity = parseNode.getObjectValue(
+      createTestEntityFromDiscriminator
+    ) as TestEntity;
+    assert.isNotNull(testEntity);
+    assert.isUndefined(testEntity.officeLocation);
```
gavinbarron, line 31: Should this be undefined or null?
I ask as undefined can be the result of the property existing and being set as undefined or the property itself not existing.
The null case can only every happen if the property exists.
If undefined is the right value, do we need an addition assertion on `testEntity.additionalData.hasOwnProperty("mobilePhone")`?

packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -0,0 +1,168 @@
...
+  private depth = -1;
+  public onBeforeObjectSerialization: ((value: Parsable) => void) | undefined;
+  public onAfterObjectSerialization: ((value: Parsable) => void) | undefined;
+  public onStartObjectSerialization:
+    | ((value: Parsable, writer: SerializationWriter) => void)
+    | undefined;
+  public writeStringValue = (key?: string, value?: string): void => {
+    key && value && this.writePropertyName(key);
+    value && this.writer.push(`=${encodeURIComponent(value)}`);
+    key && value && this.writer.push(FormSerializationWriter.propertySeparator);
```
baywet, line 22:
```suggestion
    if (key && value) {
      this.writePropertyName(key);
      this.writer.push(`=${encodeURIComponent(value)}`);
      this.writer.push(FormSerializationWriter.propertySeparator);
    }
```

packages/serialization/form/src/formSerializationWriter.ts
baywet: thanks for calling this out, solved.

packages/serialization/form/test/common/formParseNode.ts
baywet: We're only returning undefined, never null, throughout the kiota core libs, to avoid confusion. I'm not against reviewing that assumption, but I think this should be done separately from this effort.

gavinbarron: So, are we returning undefined here because the property exists and is assigned the undefined value?

gavinbarron: Actually, I don't think it has any material difference. I was thinking about the case where you want to delete the property on the server by sending null, which is only relevant when you're mutating the deserialized object and serializing it back over the wire to the server.
I'm good here, thanks.

baywet: yeah this is to parse null values (which we do in JSON and TEXT) and not return "null" (string value)

packages/serialization/form/test/browser/index.ts
gavinbarron: In that case wouldn't we want the full test suite re-exported from this barrel?

gavinbarron: Looking into the config explains it, this file is the entry point for rollup which is used to build the bundle for the karma browser tests. So yes, I think we need the other test files to be re-exported in this file

baywet: it looks like nothing needs to be changed? resolving.

gavinbarron, line 1: @baywet don't we need the other tests to be included?
```suggestion
export * from "../common/formParseNode";
export * from "../common/formParseNodeFactory";
export * from "../common/formSerializationWriter";
export * from "../common/formSerializationWriterFactory";
```

gavinbarron (changes requested):
I think that we need to make sure that we have all the tests getting run in the browser context, other that that this looks good to me

gavinbarron (approved):
LGTM!

PR #485, opened 2023-01-05
Modifying serializers as JS functions, models as interfaces

nikithauc:


packages/serialization/form/test/browser/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from "../common/formParseNode";
-export * from "../common/formParseNodeFactory";
-export * from "../common/formSerializationWriter";
-export * from "../common/formSerializationWriterFactory";
+// export * from "../common/formParseNode";
```
baywet, line 1: commented out code

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -51,23 +50,44 @@ export class JsonParseNode implements ParseNode {
...
   };
+
   public getCollectionOfObjectValues = <T extends Parsable>(
-    type: ParsableFactory<T>
+    method: DeserializeMethod<T>
   ): T[] | undefined => {
     return (this._jsonNode as unknown[])
       .map((x) => new JsonParseNode(x))
-      .map((x) => x.getObjectValue<T>(type));
+      .map((x) => x.getObjectValue<T>(method)); // test this
```
baywet, line 59: leaving a comment to identify the todo here

baywet (commented):
Thanks for starting the work here! Since this is still in draft, I've only sampled the PR for now.
This is great progress. My main concern at this point is we've lost the discriminator capability (effectively the mapping between an odata type and the right deserialization method). I think adding back (and updating) that indirection layer would solve some of the issues you've outlined in your comments. Let me know if you want me to develop more on that aspect.

nikithauc:
> Thanks for starting the work here! Since this is still in draft, I've only sampled the PR for now. This is great progress. My main concern at this point is we've lost the discriminator capability (effectively the mapping between an odata type and the right deserialization method). I think adding back (and updating) that indirection layer would solve some of the issues you've outlined in your comments. Let me know if you want me to develop more on that aspect.

Yes, can you give an example of this? 

Why do you need this exactly? 
Each RequestBuilder action is tied to specific model types, and serializer/deserializer method.

baywet:
> > Thanks for starting the work here! Since this is still in draft, I've only sampled the PR for now. This is great progress. My main concern at this point is we've lost the discriminator capability (effectively the mapping between an odata type and the right deserialization method). I think adding back (and updating) that indirection layer would solve some of the issues you've outlined in your comments. Let me know if you want me to develop more on that aspect.
> 
> Yes, can you give an example of this?
> 
> Why do you need this exactly? Each RequestBuilder action is tied to specific model types, and serializer/deserializer method.

Let me expand on this.

If you GET /groups/id/members, this describes returning `Collection(directoryObject)` in OData terms. The returned type will be a `DirectoryObjectCollectionResponse`, with a `value` property of type `DirectoryObject[]`.
In reality, this collection will never contain directory objects, it'll contain `User`, `ServicePrincipal` and other types that derive from directory object. Each of those types have properties that need to be deserialized properly.
We can tell that an entry is a user or service principal because the service returns an `@odata.type` property (called instance annotation in OData linguo), that contains something like `#microsoft.graph.User`.
We can use that value to map to the right deserialization method, this is effectively what the discriminator method `createXXXFromDiscriminatorValue` is for.

here is roughly what we had today (before your current changes)

```TypeScript
export function createDirectoryObjectFromDiscriminatorValue(parseNode: ParseNode | undefined) : User {
    if(!parseNode) throw new Error("parseNode cannot be undefined");
    const discriminatorProperty = parseNode.getChildValue("@odata.type");
    const discriminatorValue = discriminatorProperty.getStringValue();
    switch (discriminatorValue) {
         case '#microsoft.graph.User': return new User();
     // other derived types from directory object
    }
    return new DirectoryObject();
}
```

Notice how this news up the right type? I think this needs to be updated to return the reference to the right deserialization method instead. 
And we need to roll back the change to keep passing this method, instead of the deserialization method directly.

I hope that clarifies the point, don't hesitate if you have further comments/questions.

nikithauc:
> > > Thanks for starting the work here! Since this is still in draft, I've only sampled the PR for now. This is great progress. My main concern at this point is we've lost the discriminator capability (effectively the mapping between an odata type and the right deserialization method). I think adding back (and updating) that indirection layer would solve some of the issues you've outlined in your comments. Let me know if you want me to develop more on that aspect.
> > 
> > 
> > Yes, can you give an example of this?
> > Why do you need this exactly? Each RequestBuilder action is tied to specific model types, and serializer/deserializer method.
> 
> Let me expand on this.
> 
> If you GET /groups/id/members, this describes returning `Collection(directoryObject)` in OData terms. The returned type will be a `DirectoryObjectCollectionResponse`, with a `value` property of type `DirectoryObject[]`. In reality, this collection will never contain directory objects, it'll contain `User`, `ServicePrincipal` and other types that derive from directory object. Each of those types have properties that need to be deserialized properly. We can tell that an entry is a user or service principal because the service returns an `@odata.type` property (called instance annotation in OData linguo), that contains something like `#microsoft.graph.User`. We can use that value to map to the right deserialization method, this is effectively what the discriminator method `createXXXFromDiscriminatorValue` is for.
> 
> here is roughly what we had today (before your current changes)
> 
> ```ts
> export function createDirectoryObjectFromDiscriminatorValue(parseNode: ParseNode | undefined) : User {
>     if(!parseNode) throw new Error("parseNode cannot be undefined");
>     const discriminatorProperty = parseNode.getChildValue("@odata.type");
>     const discriminatorValue = discriminatorProperty.getStringValue();
>     switch (discriminatorValue) {
>          case '#microsoft.graph.User': return new User();
>      // other derived types from directory object
>     }
>     return new DirectoryObject();
> }
> ```
> 
> Notice how this news up the right type? I think this needs to be updated to return the reference to the right deserialization method instead. And we need to roll back the change to keep passing this method, instead of the deserialization method directly.
> 
> I hope that clarifies the point, don't hesitate if you have further comments/questions.

I changed the `ParsableFactory` to return  the `deserializer function of the model` instead of the model instance.

packages/serialization/form/src/formParseNode.ts
```diff
@@ -1,5 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
gavinbarron, line 1: Why is this rule being disabled for the file?

packages/serialization/form/test/common/formParseNode.ts
```diff
@@ -24,15 +24,12 @@ describe("FormParseNode", () => {
...
-    ) as TestEntity;
+    const testEntity = parseNode.getObjectValue(deserializeTestEntity);
     assert.isNotNull(testEntity);
     assert.isUndefined(testEntity.officeLocation);
     assert.equal(testEntity.id, "48d31887-5fad-4d73-a9f5-3c356e68a038");
-    assert.containsAllKeys(testEntity.additionalData, ["jobTitle"]);
-    assert.doesNotHaveAllKeys(testEntity.additionalData, ["mobilePhone"]);
-    assert.equal(testEntity.additionalData.jobTitle, "Auditor");
+    // assert.containsAllKeys(testEntity.additionalData, ["jobTitle"]);
+    // assert.doesNotHaveAllKeys(testEntity.additionalData, ["mobilePhone"]);
```
gavinbarron, line 32:
```suggestion
    assert.equal(testEntity["jobTitle"], <expectedTitle>);
    assert.equal(testEntity.hasOwnProperty("mobilePhone"), false);
```

packages/serialization/text/src/textParseNode.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
gavinbarron, line 1: We probably shouldn't be disabling this rule for the whole file

packages/test/generatedCode/apiClient.ts
```diff
@@ -1,20 +1,25 @@
...
 import {enableBackingStoreForSerializationWriterFactory, getPathParameters, ParseNodeFactoryRegistry, registerDefaultDeserializer, registerDefaultSerializer, RequestAdapter, SerializationWriterFactoryRegistry} from '@microsoft/kiota-abstractions';
+import {FormParseNodeFactory, FormSerializationWriterFactory} from '@microsoft/kiota-serialization-form';
 import {JsonParseNodeFactory, JsonSerializationWriterFactory} from '@microsoft/kiota-serialization-json';
 import {TextParseNodeFactory, TextSerializationWriterFactory} from '@microsoft/kiota-serialization-text';
 
 /**
  * The main entry point of the SDK, exposes the configuration and the fluent API.
  */
 export class ApiClient {
+    /** Path parameters for the request */
```
gavinbarron, line 12: somehow the comments here are doubled up

gavinbarron:
The build here are showing nicely the intermittent generation issue with case and package: https://github.com/microsoft/kiota-typescript/actions/runs/3999492551/jobs/6863470231

packages/serialization/form/test/common/formSerializationWriter.ts
```diff
@@ -40,6 +40,12 @@ describe("FormSerializationWriter", () => {
       "accountEnabled=false&" +
       "jobTitle=Author&" +
       "createdDateTime=1970-01-01T00%3A00%3A00.000Z";
+      console.log("form:");
+      console.log(form);
+      console.log("form:");
+      console.log("expectedString:");
+      console.log(expectedString);
+      console.log("expectedString:");
```
baywet, line 48:
```suggestion
```

baywet (changes requested):

packages/serialization/text/src/textParseNode.ts
nikithauc: Done

packages/test/generatedCode/models/serializeAttachmentCollectionResponse.ts
```diff
@@ -3,6 +3,17 @@ import {serializeAttachment} from './serializeAttachment';
 import {AdditionalDataHolder, DeserializeIntoModelFunction, Parsable, ParseNode, SerializationWriter} from '@microsoft/kiota-abstractions';
 
 export function serializeAttachmentCollectionResponse(writer: SerializationWriter, attachmentCollectionResponse: AttachmentCollectionResponse | undefined = {}) : void {
-            writer.writeStringValue("@odata.nextLink", attachmentCollectionResponse.odataNextLink);
-            writer.writeCollectionOfObjectValues<Attachment>("value", attachmentCollectionResponse.value, serializeAttachment);
+        for (const [key, value] of Object.entries(attachmentCollectionResponse)){
+            switch(key){
+                case "@odata.nextLink":
```
baywet, line 8: this should be the symbol name and not the serialization name being used here

packages/serialization/form/test/common/formSerializationWriter.ts
```diff
@@ -31,22 +31,33 @@ describe("FormSerializationWriter", () => {
...
+        count++;
+      }
+    });
+    assert.equal(expectedString.length, count);
+    assert.equal(arr.length, 0);
+   
+    console.log("expectedString:");
+    console.log(expectedString);
+    console.log("expectedString:");
+    // assert.equal(form, expectedString);
```
baywet: commented out code

baywet (changes requested):

packages/test/generatedCode/models/serializeAttachmentCollectionResponse.ts
nikithauc: done

baywet (changes requested):
Getting really close, just a single comment on my end

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -51,23 +50,46 @@ export class JsonParseNode implements ParseNode {
...
       .map((x) => new JsonParseNode(x))
-      .map((x) => x.getObjectValue<T>(type));
+      .map((x) => x.getObjectValue<T>(method));
   };
-  public getObjectValue = <T extends Parsable>(type: ParsableFactory<T>): T => {
-    const result = type(this);
+
+  public getObjectValue = <T extends Parsable>(
+    parsableFactory: ParsableFactory<T>,
+    value: T = {} as T
```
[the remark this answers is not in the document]
nikithauc: Since the parsable factory returns the deserializer and not an instance, we need to pass a value which is  a model object exclusively which can be used to set the result.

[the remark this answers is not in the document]
baywet: @koros the prototype of this method still needs to be updated here

packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -118,7 +118,7 @@ export class JsonSerializationWriter implements SerializationWriter {
 
     this.onStartObjectSerialization &&
       this.onStartObjectSerialization(value as unknown as Parsable, this);
-    value && serializerMethod && serializerMethod(this, value);
+    value && serializerMethod(this, value);
```
baywet, line 121: no sure about this change, we want to test whether the method reference has been passed before calling it.

koros: I agree, I have reverted the change

baywet (dismissed):

gavinbarron (changes requested):
Just one question around co-ordination of merging between this repo and the generator repo.

Otherwise this looks great to me!

baywet (approved):

baywet (dismissed):

andrueastman (approved):

.github/workflows/generate_graph_test.yml
```diff
@@ -16,11 +16,12 @@ on:
...
 
   push:
-    branches: [ main, nikithauc/model-interface-serialization ]
+    branches: [ main ]
   pull_request:
+    branches: [main]
 
 env:
-  KIOTA_BRANCH: 'main'
+  KIOTA_BRANCH: 'nikithauc/model-as-interfaces'
```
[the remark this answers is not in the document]
baywet:
```suggestion
  KIOTA_BRANCH: 'main'
```

baywet (approved):

PR #488, opened 2023-01-06
feature/user agent

baywet:
related https://github.com/microsoft/kiota/issues/2056

### DIFF
--- a/packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
+++ b/packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
@@ -7,11 +7,11 @@
 
 import { RequestOption } from "@microsoft/kiota-abstractions";
 
-export const UserAgentHandlerOptionKey = "UserAgentHandlerOptionKey";
+export const UserAgentHandlerOptionsKey = "UserAgentHandlerOptionKey";
 
-export class UserAgentHandlerOption implements RequestOption {
+export class UserAgentHandlerOptions implements RequestOption {
 	getKey(): string {
-		return UserAgentHandlerOptionKey;
+		return UserAgentHandlerOptionsKey;
 	}
 	/**
 	 * @public
--- a/packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
+++ b/packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
@@ -7,6 +7,8 @@
 
 import { RequestOption } from "@microsoft/kiota-abstractions";
 
+import { libraryVersion } from "./version";
+
 export const UserAgentHandlerOptionsKey = "UserAgentHandlerOptionKey";
 
 export class UserAgentHandlerOptions implements RequestOption {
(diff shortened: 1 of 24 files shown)


packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
```diff
@@ -0,0 +1,25 @@
...
+	}
+	/**
+	 * @public
+	 * @constructor
+	 * To create an instance of UserAgentHandlerOption
+	 * @param {boolean} [enable = true] - Whether to add the user agent header to the request
+	 * @param {string} [productName = "kiota-typescript"] - The product name to be added to the user agent header
+	 * @param {string} [productVersion = "1.0.0-preview.12"] - The product version to be added to the user agent header
+	 */
+	public constructor(public enable: boolean = true, public productName: string = "kiota-typescript", public productVersion: string = "1.0.0-preview.12") {}
```
gavinbarron, line 24: This feels prone to falling out of date.
Can write the version from package.json into a constant on build and use that value?
We have a process that does this during build for mgt

packages/http/fetch/test/common/middleware/userAgentHandler.ts
```diff
@@ -0,0 +1,47 @@
...
+			const options = new UserAgentHandlerOptions();
+			const handler = new UserAgentHandler(options);
+			handler.next = dummyFetchHandler;
+			dummyFetchHandler.setResponses([new Response("ok", { status: 200 }), new Response("ok", { status: 200 })]);
+			const requestUrl = "url";
+			const fetchRequestInit = {
+				method: "GET",
+			};
+			await handler.execute(requestUrl, fetchRequestInit, undefined);
+			await handler.execute(requestUrl, fetchRequestInit, undefined);
```
gavinbarron, line 37: Why the duplicated call here?

baywet: to verify idempotance :)

packages/http/fetch/src/middlewares/options/userAgentHandlerOptions.ts
baywet: sure, do you have pointers to share?

baywet: nevermind, just pushed a script, thanks for calling this out.

baywet: ah, I hate that GitHub doesn't refresh the comments live. Didn't want to put my hands in gulp, I figured a powershell script would do the trick.

scripts/updateVersion.ps1
```diff
@@ -0,0 +1,27 @@
...
+}
+
+if (-not (Test-Path $versionFilePath)) {
+	Write-Error "Could not find version file at path: $versionFilePath"
+	Exit 1
+}
+
+$packageDefinition = Get-Content $packageDefinitionPath | ConvertFrom-Json
+$version = $packageDefinition.version
+Set-Content $versionFilePath -Value "export const libraryVersion = ""$version"";" -Verbose
```
gavinbarron, line 27: very nit-picky, but perhaps include the copyright header in the generated file so that it's present when the file is checked in.

gavinbarron: Other than it's recommended that we have it in internal guidance, not much at all.

gavinbarron (approved):

PR #493, opened 2023-01-09
ci: adding automation for end to end generation testing

gavinbarron:


nikithauc (approved):

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,81 @@
...
+        - name: Archive dist folders # archive dist folders to verify if they are transpiled correctly and available for publishing
+          uses: actions/upload-artifact@v3
+          with:
+            name: dist folders
+            path: |
+              packages/abstractions/dist
+              packages/serialization/json/dist
+              packages/serialization/text/dist
+              packages/http/fetch/dist
+              packages/authentication/azure/dist
```
baywet, line 74: do we care about those dists as we're not going to publish from this workflow? Can we instead upload a webpack bundle analyzer report to have an history of the changes?

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,81 @@
...
+        description: 'OpenAPI definition to build test SDK for'
+        type: string
+        default: 'https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/openapi/v1.0/openapi.yaml'
+
+  push:
+    branches: [ nikithauc/model-interface-serialization ]
+  pull_request:
+
+env:
+  KIOTA_BRANCH: 'nikithauc/model-as-interfaces'
```
baywet, line 23:
```suggestion
  KIOTA_BRANCH: ${{ inputs.KIOTA_BRANCH }}
```

can you define it this way instead? so you don't have to read both in the workflow?

baywet (changes requested):

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,81 @@
...
+        type: string
+        default: 'nikithauc/model-as-interfaces'
+      KIOTA_REPO:
+        description: 'Repo to build kiota from'
+        type: string
+        default: 'https://github.com/microsoft/kiota.git'
+      OPEN_API_DEF:
+        description: 'OpenAPI definition to build test SDK for'
+        type: string
+        default: 'https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/openapi/v1.0/openapi.yaml'
```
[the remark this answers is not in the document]
nikithauc: I believe the intention is to test the Graph surface and keep an eye on the package size and number of files.

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,81 @@
...
+          if: ${{ inputs.KIOTA_BRANCH }}
+        - run: ./kiota-linux-x64/kiota generate --openapi ${{ env.OPEN_API_DEF }} -l typescript --output ./packages/test/generatedCode --clean-output
+          if: ${{ !inputs.KIOTA_BRANCH }}
+        - uses: actions/checkout@v3
+        - name: Use Node.js ${{ matrix.node-version }}
+          uses: actions/setup-node@v3
+          with:
+            node-version: ${{ matrix.node-version }}
+        - run: yarn install --frozen-lockfile --ignore-engines #spfx still not compatible with node 18
+        - run: yarn global add mocha@10.X
```
nikithauc, line 64:
```suggestion
        - run: yarn global add mocha@10.X
        - run: yarn run build
```

.github/workflows/generate_graph_test.yml
gavinbarron: That is the plan, I'm iterating on getting this working and then adding the functionality I want.

.github/workflows/generate_graph_test.yml
gavinbarron: I was finding that when it was triggered from a PR the inputs were not set so I added the env as a fallback.

.github/workflows/generate_graph_test.yml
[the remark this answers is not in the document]
gavinbarron: Yes, the idea here is to have a means of tracking the size of the full graph sdk build.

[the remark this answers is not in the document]
baywet: alright, thanks for the precision here. resolving.

.github/workflows/generate_graph_test.yml
gavinbarron: So, because we don't build the SDK with webpack I don't think that we can use the webpack bundle analyzer for that aspect, but I've used `npm pack --dry-run --json` instead to get some package size data and output it as a PR comment.

My thinking is that I'll also add a small sample application which is built using webpack that uses the generated SDK and that way we can also track how we're doing on the bundle size for a client app and that would use WebPack.

baywet:
@gavinbarron it'd probably have been easier to rebase Nikitha's branch onto main and then this onto her branch to avoid tracking changes made to main in that PR.
That or merging main to her branch and then rebasing this onto it. Thoughts?

gavinbarron:
Yeah, there's definitely a need to have this work with both of Nikitha's working branches. 
Updating her working branch from main makes the most sense to me and I'll work with @nikithauc to get things into a good state.

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,107 @@
...
+  KIOTA_REPO: 'https://github.com/microsoft/kiota.git'
+  OPEN_API_DEF: 'https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/openapi/v1.0/openapi.yaml'
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        environment:
+          name: build_test
+        env:
+          TENANT_ID: ${{ secrets.TENANT_ID }}
```
baywet, line 33:
```suggestion
          TENANT_ID: ${{ secrets.TENANT_ID }}
          NODE_OPTIONS: "--max_old_space_size=9182"
```

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,107 @@
...
+          if: ${{ inputs.KIOTA_BRANCH }}
+        - run: ./kiota-linux-x64/kiota generate --openapi ${{ env.OPEN_API_DEF }} -l typescript --output packages/test/generatedCode --clean-output --log-level debug
+          if: ${{ !inputs.KIOTA_BRANCH }}
+
+        - run: yarn install --frozen-lockfile --ignore-engines #spfx still not compatible with node 18
+        - run: yarn global add mocha@10.X
+        - run: yarn tsc:version
+        - name: yarn build
+          env:
+            NODE_OPTIONS: "--max_old_space_size=9182"
```
baywet, line 69:
```suggestion
```

.github/workflows/generate_graph_test.yml
```diff
@@ -0,0 +1,107 @@
...
+          run: yarn build
+        # build the sample generated SDK code only
+        - name: yarn build:sdk
+          env:
+            NODE_OPTIONS: "--max_old_space_size=9182"
```
baywet, line 74:
```suggestion
```

baywet (approved):
minor suggestion for the space size

baywet:
@gavinbarron converted to draft to help me sort through the PRs I need to review if you don't mind :)

.github/workflows/generate_graph_test.yml
gavinbarron: Marking this as resolved now that there is a sample app which we can report on the bundle size with.

gavinbarron:
@baywet published as I'm now happy with the build and outputs.

nikithauc (dismissed):

baywet (changes requested):
suggesting conditions so the workflow doesn't fail on other triggers

gavinbarron:
Thanks @baywet very good point!

baywet (approved):
thanks for taking the changes in

PR #520, opened 2023-02-06
Selective imports

nikithauc:
Using modular augmentation and `Reflect.defineProperty` to achieve selective imports for each path segment.  

I recommend playing with this and feeling the developer experience with selective imports. 

One thing to remember, typescript code is transpiled into a `outDir`, let's say `dist`. The main and types entry point is set to `dist/index. js/ts`. But since we are not exporting everything from a single barrel the user would probably have to something like

```
import "@msgraphsdk/dist/users"
import "@msgraphsdk/dist/users/item"
```

[Re-exporting the index files](https://github.com/sebastienlevert/kiota-typescript/commit/797ff17ff3a126284452d3a3133aaf13a3f90eaa#diff-8f05766dd02329f2edffb772760fc9b70d5a44a6e6e20bd783c1270f30de501c)  to have a single point of entry for all the index files will bring in the `Reflect.define` for every import and will pull in all code not allowing treeshaking.


packages/test/generatedCode/apiClient.ts
```diff
@@ -1,57 +1,43 @@
...
-        return new UserItemRequestBuilder(urlTplParams, this.requestAdapter);
-    };
-}
+                /**
+                 * The main entry point of the SDK, exposes the configuration and the fluent API.
+                 */
+                export class ApiClient {
+                    /** Path parameters for the request */
+                    /** Path parameters for the request */
+                    public pathParameters: Record<string, unknown>;
```
baywet, line 14: I don't believe the change in visibility is a change we wanted here., Can your check for regressions in kiota changes please?

packages/test/generatedCode/apiClient.ts
```diff
@@ -1,57 +1,43 @@
...
-        const urlTplParams = getPathParameters(this.pathParameters);
-        urlTplParams["user%2Did"] = id
-        return new UserItemRequestBuilder(urlTplParams, this.requestAdapter);
-    };
-}
+                /**
+                 * The main entry point of the SDK, exposes the configuration and the fluent API.
+                 */
+                export class ApiClient {
+                    /** Path parameters for the request */
```
baywet, line 12: noting that here knowing it may come from the parent PR, but the doc comments are being inserted twice.

packages/test/generatedCode/users/item/index.ts
```diff
@@ -0,0 +1,19 @@
+    import {UserItemRequestBuilder } from "./userItemRequestBuilder"
+    import {ApiClient} from "../../apiClient"
+    import { getPathParameters } from "@microsoft/kiota-abstractions";
+    declare module "../../apiClient"{
+        interface ApiClient{
+            usersById:(id : string) => UserItemRequestBuilder
```
baywet, line 6: maybe it should have a doc comment?

baywet (changes requested):
great progress on this front! a couple of comments to help drive this home

PR #525, opened 2023-02-07
- removes nikitha from the code owners

baywet:
Good luck on the next team! Thanks for all the contributions!


sebastienlevert:
Just want to mention that the output size of the app is impressive! The app is using selective imports I assume?

baywet:
@sebastienlevert not on this branch, this branch only has a tiny subset of the API surface.

gavinbarron:
The number that we're most interested in is the Parsed size: 13MB, that is a 13 MB app @sebastienlevert,
We have a long way to go to reach something good, the gzipped number shows how much real world network traffic that package would cost.

gavinbarron (approved):

PR #569, opened 2023-03-22
feature/size

baywet:
related https://github.com/microsoft/kiota/issues/2388


sebastienlevert:
I have a question here. As we're using Lerna for our mono repo, shouldn't we use Lerna features to automatically have packages reference each other with their latest versions? That would remove a lot of risks when updating 10+ packages... There was work done on the Graph Toolkit repo that solve this specific problem and I think we could get good efficiency if we embraced something like this. Not my field of expertise, but seeing so many packages updates scares me 😃

sebastienlevert:
I'm afraid this creates a supportability matrix that will be a monster to manage... But I understand your point. MGT has been proven successful with a single version... But it might be better for Kiota to have separate package versions. Something to think about.

baywet:
> Note: The package size hasn't reduced yet because I have not refreshed the generated project to avoid getting conflicts with the interface/model work.

koros (approved):

PR #619, opened 2023-04-26
Bump uri template

### LINKED ISSUE #584: `uri-template-lite` breaks when using a bundler
Trying to use `kiota-typescript` from a UI application that is bundled using `vite` we have been experiencing issues with `uri-template-lite`.

unnamed

The current workaround is to exclude the package from the bundler:

```
  build: {
    ...
    rollupOptions: {
      external: ['uri-template-lite'],
    }
  },
```

but is less than ideal and it compromises the "getting started" experience.

We have various options here:
- include the single source of `uri-template-lite` in this codebase
- rely on a different library (that better plays with modules etc.)
- open to other ideas!

andreaTP:
This should fix #584 , I couldn't verify that this completely solves the problem but is a good guess ...


edewit (dismissed):

packages/abstractions/tsconfig.base.json
```diff
@@ -13,6 +13,5 @@
         "declarationMap": true,
         "sourceMap": true,
         "composite": true,
```
baywet, line 15:
```suggestion
        "composite": true
```
I'm not sure how that ever worked since it's not valid JSON....

packages/abstractions/package.json
```diff
@@ -38,7 +38,7 @@
     "@opentelemetry/api": "^1.2.0",
     "tinyduration": "^3.2.2",
     "tslib": "^2.3.1",
-    "uri-template-lite": "^22.1.0",
+    "uri-template-lite": "23.4.0",
```
baywet, line 41: is there any specific reason why you're locking on that specific exact version as opposed to charet? (current major and anything under)

baywet (changes requested):
Thanks for the contribution as always.
We use yarn here and lerna, would you mind dropping the package lock and running a yarn install for sanity please? Also left a couple of comments.

packages/abstractions/tsconfig.base.json
andreaTP: I think it works anyhow here, but 🤷

packages/abstractions/package.json
andreaTP: Just being a bit defensive here, would you prefer a `^23.4.0` ?

baywet: correct, thanks

baywet (approved):
@koros for final review and testing in a browser setup + in a node app setup

andreaTP:
should be ready now 🙂

baywet (approved):

koros:
> @koros have you had the time to look into this PR? Anything I can do?
@andreaTP I'll check it out and get back to you

koros (dismissed):

koros:
@andreaTP The code looks okay on my end, however I am not sure why the pipeline tests failed here: https://github.com/microsoft/kiota-typescript/actions/runs/4809556938/jobs/8560999617?pr=619, also which version of kiota generator should I use to run more tests

andreaTP:
I'll rebase this early tomorrow if this is still open :+1:

koros:
> @koros , rebased on latest `main`, the CI should be green, can you please trigger it?

@andreaTP all the tests are okay now, Thanks. There is a minor issue though https://github.com/microsoft/kiota-typescript/actions/runs/5014471823/jobs/8989301442?pr=619 It looks like the CI/CD is trying to add a comment about the bundle size but it probably can't access your resource.  I can merge it if there is no other comment from @sebastienlevert @gavinbarron

sebastienlevert:
> > @koros , rebased on latest `main`, the CI should be green, can you please trigger it?
> 
> @andreaTP all the tests are okay now, Thanks. There is a minor issue though https://github.com/microsoft/kiota-typescript/actions/runs/5014471823/jobs/8989301442?pr=619 It looks like the CI/CD is trying to add a comment about the bundle size but it probably can't access your resource. I can merge it if there is no other comment from @sebastienlevert @gavinbarron

LGTM!

andreaTP:
thanks @koros for the feedback, opened #667 to track the issue with the CI.

koros (approved):

andreaTP:
@koros I'm not sure I fully understand the problem here, let me go through it.

> it looks like the CI wont allow merge before all checks

either you disable the check as proposed in #667 or you disable the check as "mandatory" in the project settings.

> in the past how were the PRs done

this is my first PR to this repository, and, apparently, this is the first from a community member that is not part of the microsoft-graph organization.

> is the issue we're encountering new or its something known.

the issue with the CI is something that happened in other repositories, e.g.:
https://github.com/microsoft/kiota-java/pull/142
https://github.com/microsoft/kiota/issues/2591

andreaTP:
Rebased on latest `main`, @koros can you, please trigger the CI?

koros:
@andreaTP there's something else broken but it has nothing to do with this PR, its the generation side let me look into that.

andreaTP:
Thanks for chipping in @MIchaelMainer !

> Folks, we shouldn't be merging directly into the branch where we publish from.

This is not a common practice, but I have no objection if someone is willing to do the manual work all of the time ...

> We should consider creating another branch off of main, create new PR into the branch, merge into that branch, and then the CI/CD can run when you create another PR into main. 

This could work, but I expect it to work now as #668 got merged.

At the moment seems like something is broken in Kiota `main` and the test is correctly failing because of it.

andreaTP:
Just to make it clear, the current CI failure doesn't seem to be related with the changes proposed in this PR.

andreaTP:
@baywet sorry for bothering you directly, would you be able to facilitate the progress on this PR?

baywet (approved):

baywet:
@andreaTP it seems our CI is impacted by this https://github.com/microsoft/kiota/issues/2676 which makes it flaky. I just reran the thing for now and it passed.
Thank you for your patience while I was away. We'll need to investigate the generation behaviour, but that's a separate issue we've already captured.
Merging.

ksdaniel:
Not sure this is related - but, with Vite and:  

```
  "version": "1.0.0-preview.19"
```

i get an exception: 

```
Uncaught (in promise) TypeError: Template is not a constructor
    at get URL [as URL] (requestInformation.ts:42:24)
    at FetchRequestAdapter.<anonymous> (fetchRequestAdapter.js:452:45)
```

Looking at the packages for kiota-abstractions: 

```
  "dependencies": {
    "@opentelemetry/api": "^1.2.0",
    "tinyduration": "^3.2.2",
    "tslib": "^2.3.1",
    "uri-template-lite": "^23.4.0",
    "uuid": "^9.0.0"
  },
```

we are on the new version of uri-template-lite. 

But it code seem to break here: 

```
 const template = new Template(this.urlTemplate);
          const data = {};
``` 

aka: 

```
      const template = new (Template as any)(this.urlTemplate);
``` 

Got any ideas? This is a rather "blocking" issue.

baywet:
Thanks for reporting this. Do you have the same issue if you downgrade the package?

ksdaniel:
If I downgrade to 

```
    "@microsoft/kiota-abstractions": "1.0.0-preview.18",
    "@microsoft/kiota-http-fetchlibrary": "^1.0.0-preview.18",
    "@microsoft/kiota-serialization-form": "^1.0.0-preview.8",
    "@microsoft/kiota-serialization-json": "^1.0.0-preview.17",
    "@microsoft/kiota-serialization-text": "^1.0.0-preview.16",
``` 

I no longer see the error, mentioned before. (I only downgraded kiota-abstractions)

But now there seems to be a mismatch between kiota-abstractions and kiota-http-fetchlibrary and I get different errors. 

Any idea what the working combination might be?

ksdaniel:
Just an update - in the end, this combination ran for me: 

```
    "@microsoft/kiota-abstractions": "1.0.0-preview.18",
    "@microsoft/kiota-http-fetchlibrary": "1.0.0-preview.17",
    "@microsoft/kiota-serialization-form": "1.0.0-preview.7",
    "@microsoft/kiota-serialization-json": "1.0.0-preview.16",
    "@microsoft/kiota-serialization-text": "1.0.0-preview.15",
```

Able to compile and bundle with Vite - the definition mismatches are also gone.

andreaTP:
@ksdaniel would you be able to provide a minimal reproducer so that we can fix the underlying issue?

PR #715, opened 2023-06-19
Fix uri-template-lite binding

andreaTP:
Fix for this comment:
https://github.com/microsoft/kiota-typescript/pull/619#issuecomment-1596171382
I manually tested it with a Vanilla fresh vite build and now it works as expected.

cc. @ksdaniel would you be able to confirm?

### DIFF
--- a/packages/abstractions/types/uri-template-lite/index.d.ts
+++ b/packages/abstractions/types/uri-template-lite/index.d.ts
@@ -1,5 +1,3 @@
 declare module 'uri-template-lite' {
-  class Template {
-    constructor(template: string);
-  }
+  function expand(template: string, data: { [key: string]: unknown }): string;
 }
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -39,7 +39,6 @@ export class RequestInformation {
     } else if (!this.urlTemplate) {
       throw new Error("urlTemplate cannot be undefined");
     } else {
-      const template = new (Template as any)(this.urlTemplate);
       const data = {} as { [key: string]: unknown };
       for (const key in this.queryParameters) {
         if (this.queryParameters[key]) {
@@ -51,7 +50,7 @@ export class RequestInformation {
           data[key] = this.pathParameters[key];
         }
       }
-      return template.expand(data);
+      return Template.expand(this.urlTemplate, data);
     }
   }
   /** Sets the URL of the request */


packages/abstractions/types/uri-template-lite/index.d.ts
```diff
@@ -1,5 +1,3 @@
 declare module 'uri-template-lite' {
-  class Template {
-    constructor(template: string);
-  }
+  function expand(template: string, data: { [key: string]: unknown }): string;
```
baywet, line 2: @andreaTP is there any specific reasons why we're not using the type definitions here?
https://www.npmjs.com/package/@types/uri-template-lite
Especially since the author of the type definitions is a Microsoft employee who works on Microsoft Graph and we can reach out if we need help.

andreaTP:
the CI failure seems to track a kiota CLI regression, or should I fix anything here?

packages/abstractions/types/uri-template-lite/index.d.ts
baywet: Besides the version comment, what makes you say they are outdated? especially in the context of using a single function.

For type definitions on DefinitelyTyped, you can just PR them, they have a bot that brings past contributors to review them. If you don't get traction let us know, I'll ping Vincenzo internally.

ksdaniel:
> 

Using your branch, I no longer see the error.

baywet:
> the CI failure seems to track a kiota CLI regression, or should I fix anything here?

@andreaTP this is still the flaky issue we need to address in the generator. Staffing has been challenging for over a year now and we appreciate your patience. I'll override the policies if we need again. Can you have a look at using the typings instead of the custom ones please?

packages/abstractions/types/uri-template-lite/index.d.ts
andreaTP: as the user noticed this constructor is not working as expected after roll-up:

```ts
    class Template {
        constructor(template: string);
    }
```

probably because `Template` is a `function`: https://github.com/litejs/uri-template-lite/blob/654515a40bd0f01141119896f7f2b4e9cb23c106/index.js#L89

I'm not 100% sure how to properly fix the upstream types ...
I also think that the burden to keep this file in the repo is pretty low, and gives us shorter feedback loops when something changes.

andreaTP:
> this is still the flaky issue we need to address in the generator.

got it, no worries, do you need any help fixing it?

> Can you have a look at using the typings instead of the custom ones please?

I did, but unfortunately, I feel that keeping the type locally to this project is still the best option.
That said, if you have any TS expert that knows the answer on top of their head I'm happy to follow the lead!

packages/abstractions/types/uri-template-lite/index.d.ts
baywet: To be specific I'm not suggesting you change the code in this PR for RequestInformation. Just that you try to strip the types definitions from here to replace them with the @types package. The less code we have to maintain the better :)

andreaTP: I completely understand, but I cannot find a working encoding of the exported `Template` ... it's a very very javascript thing ...

baywet: wouldn't `import * as URI from 'uri-template-lite';` and then `URI.expand(this.urlTemplate, data);` work once this is removes and the types are added?

andreaTP: Attempted a bunch of variations, in this case:

```
> @microsoft/kiota-abstractions@1.0.0-preview.19 build:cjs
> tsc -p tsconfig.cjs.json

src/requestInformation.ts:54:18 - error TS2339: Property 'expand' does not exist on type 'typeof import("/Users/aperuffo/workspace/kiota-typescript/node_modules/@types/uri-template-lite/index")'.

54       return URI.expand(this.urlTemplate, data);
                    ~~~~~~


Found 1 error in src/requestInformation.ts:54
```

andreaTP: Note, in that case the following TS compiles:

```ts
URI.URI.expand(this.urlTemplate, data)
```

but it doesn't work at runtime in a vite build at least.

andreaTP: Thanks a lot for the detailed research!
After this I'm more convinced that keeping this definition locally is the best course of action. :-) If you don't mind...

baywet (approved):

Issue #40, opened 2021-08-03
typescript core (and generation) should be using UUID


baywet:
https://www.npmjs.com/package/uuid or https://www.npmjs.com/package/guid-typescript

https://github.com/microsoft/kiota/blob/47f57ce64af7856d18711f1895dac48d3f1048ae/abstractions/typescript/src/serialization/parseNode.ts#L32
https://github.com/microsoft/kiota/blob/47f57ce64af7856d18711f1895dac48d3f1048ae/serialization/typescript/json/src/jsonParseNode.ts#L16
https://github.com/microsoft/kiota/blob/47f57ce64af7856d18711f1895dac48d3f1048ae/src/Kiota.Builder/Writers/TypeScript/TypeScriptConventionService.cs#L85 
[AB#10487](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10487)

baywet:
Ping @nikithauc and @gavinbarron on the matter as I ran into that today again.
guid-typescript is clearly not maintained anymore. There are other alternatives out there.
Currently mapping guids to strings, which is probably subpar.

koros:
Fixed by https://github.com/microsoft/kiota-typescript/pull/745

PR #745, opened 2023-07-10
handle Guid types in typescript

koros:


packages/http/fetch/package.json
```diff
@@ -40,6 +40,7 @@
 	"dependencies": {
 		"@microsoft/kiota-abstractions": "^1.0.0-preview.22",
 		"@opentelemetry/api": "^1.2.0",
+		"guid-typescript": "^1.0.9",
```
andrueastman, line 43: Out of curiosity, won't this be added as an indirect dependency from the abstractions package. If so, do we need to add it here? and why not add to the other packages as well(serialization-form, serialization-json)?

packages/abstractions/package.json
```diff
@@ -36,6 +36,7 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.2.0",
+    "guid-typescript": "^1.0.9",
```
andrueastman, line 39: Just to confirm, shouldn't we bump up the package versions here so that this builds after release?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -20,7 +21,7 @@ export class JsonParseNode implements ParseNode {
     new JsonParseNode((this._jsonNode as any)[identifier]);
   public getBooleanValue = () => this._jsonNode as boolean;
   public getNumberValue = () => this._jsonNode as number;
-  public getGuidValue = () => this._jsonNode as string;
+  public getGuidValue = () => Guid.parse(this.getStringValue());
```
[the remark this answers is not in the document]
andrueastman: https://github.com/snico-dev/guid-typescript/blob/f110c1f42b5c784511e322531e97928c2d8ffb1c/lib/guid.ts#L40

packages/http/fetch/package.json
koros: Yes you're right ill remove it

packages/abstractions/package.json
baywet: yes we should, but only at the last minute since the tags will be what gets published

packages/http/fetch/package.json
baywet: actually in typescript/javascript land, if your code has a direct dependency on a package (i.e. you have an import pointing to that package in your code), it should be declared as a dependency. Granted we control abstractions as well, but I learnt that the hard way...

baywet (dismissed):

packages/http/fetch/package.json
```diff
@@ -38,7 +38,7 @@
 		"test:cjs": "mocha 'dist/cjs/test/common/**/*.js' && mocha  'dist/cjs/test/node/**/*.js'"
 	},
```
baywet, line 39: the guid package should be a dev dependency here

packages/serialization/json/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "homepage": "https://github.com/microsoft/kiota-typescript#readme",
   "dependencies": {
```
baywet, line 41: guid package should be a dependency here

packages/serialization/text/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "homepage": "https://github.com/microsoft/kiota-typescript#readme",
   "dependencies": {
-    "@microsoft/kiota-abstractions": "^1.0.0-preview.22",
```
baywet, line 42: guid package should be a dependency here

packages/serialization/form/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "homepage": "https://github.com/microsoft/kiota-typescript#readme",
   "dependencies": {
```
baywet, line 41: guid package should be a dependency here

baywet (changes requested):

packages/http/fetch/package.json
koros: @baywet does it means the documentation will have to be updated once we merge because the generated code, in certain cases, now contains `import { Guid } from "guid-typescript";`

baywet (approved):

andrueastman (approved):

koros:
fixes https://github.com/microsoft/kiota-typescript/issues/40

PR #752, opened 2023-07-12
fix: fixed enum parsing for json data

### LINKED ISSUE #744: [BUG]: Kiota fails to deserialize response when it contains array of enums
Kiota fails to de-serialize the response when it contains this array field of enum in OpenAPI:

```yaml
xyz:
  type: array
  items:
    type: string
    enum:
      - foo
      - bar
      - baz
```

It generates code like this:
```js
"xyz": n => { instance.xyz = n.getEnumValues(xyz); }
```

It calls [getEnumValues method](https://github.com/microsoft/kiota-typescript/blob/main/packages/serialization/json/src/jsonParseNode.ts#L94-L100) of @microsoft/kiota-serialization-json package

and it errors on

```ts
return rawValues.split(",").map((x) => type[toFirstCharacterUpper(x)] as T);
//              ^ Here, as the rawValues is already an array and array doesn't have split method
```

salmannotkhan:
refactor: updated method name `getEnumValues` to `getCollectionOfEnumValues`

Resolves #744

### DIFF
--- a/packages/serialization/json/src/jsonParseNode.ts
+++ b/packages/serialization/json/src/jsonParseNode.ts
@@ -92,19 +92,19 @@ export class JsonParseNode implements ParseNode {
       }
     });
   };
-  public getEnumValues = <T>(type: any): T[] => {
-    const rawValues = this.getStringValue();
-    if (!rawValues) {
-      return [];
+  public getCollectionOfEnumValues = <T>(type: any): T[] => {
+    if (Array.isArray(this._jsonNode)) {
+      const values = this._jsonNode.map(x => {
+        const node = new JsonParseNode(x)
+        return node.getEnumValue(type) as T
+      })
+      return values.filter(Boolean)
     }
-    return rawValues.split(",").map((x) => type[toFirstCharacterUpper(x)] as T);
+    return []
   };
   public getEnumValue = <T>(type: any): T | undefined => {
-    const values = this.getEnumValues(type);
-    if (values.length > 0) {
-      return values[0] as T;
-    } else {
-      return undefined;
-    }
+    const rawValue = this.getStringValue()
+    const value = type[toFirstCharacterUpper(rawValue)]
+    return value
   };
 }
--- a/packages/serialization/json/src/jsonParseNode.ts
+++ b/packages/serialization/json/src/jsonParseNode.ts
@@ -94,17 +94,18 @@ export class JsonParseNode implements ParseNode {
   };
   public getCollectionOfEnumValues = <T>(type: any): T[] => {
     if (Array.isArray(this._jsonNode)) {
-      const values = this._jsonNode.map(x => {
+      return this._jsonNode.map(x => {
         const node = new JsonParseNode(x)
         return node.getEnumValue(type) as T
-      })
-      return values.filter(Boolean)
+      }).filter(Boolean)
     }
     return []
   };
   public getEnumValue = <T>(type: any): T | undefined => {
     const rawValue = this.getStringValue()
-    const value = type[toFirstCharacterUpper(rawValue)]
-    return value
+    if (!rawValue) {
+      return undefined
+    }
+    return type[toFirstCharacterUpper(rawValue)]
   };
 }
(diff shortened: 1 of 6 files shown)


salmannotkhan:

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -91,19 +91,19 @@ export class JsonParseNode implements ParseNode {
...
-    const rawValues = this.getStringValue();
-    if (!rawValues) {
-      return [];
+  public getCollectionOfEnumValues = <T>(type: any): T[] => {
+    if (Array.isArray(this._jsonNode)) {
+      const values = this._jsonNode.map(x => {
+        const node = new JsonParseNode(x)
+        return node.getEnumValue(type) as T
+      })
+      return values.filter(Boolean)
```
baywet, line 100:
```suggestion
      return this._jsonNode.map(x => {
        const node = new JsonParseNode(x)
        return node.getEnumValue(type) as T
      }).filter(Boolean);
```

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -91,19 +91,19 @@ export class JsonParseNode implements ParseNode {
...
+    return []
   };
   public getEnumValue = <T>(type: any): T | undefined => {
-    const values = this.getEnumValues(type);
-    if (values.length > 0) {
-      return values[0] as T;
-    } else {
-      return undefined;
-    }
+    const rawValue = this.getStringValue()
```
baywet, line 105: this should test whether the string value is undefined or empty

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -91,19 +91,19 @@ export class JsonParseNode implements ParseNode {
...
+    const value = type[toFirstCharacterUpper(rawValue)]
+    return value
```
baywet, line 107:
```suggestion
    return type[toFirstCharacterUpper(rawValue)];
```

baywet (changes requested):
Thanks for the contribution here are a couple of suggestions.
Did you take the time to check whether the serialization writer was aligned in terms of logic as well?

salmannotkhan:
I can create a PR for that if you want

baywet:
I was referring to [this](https://github.com/microsoft/kiota-typescript/blob/78bfddff68f35306817456b0dc7d5ac922b6f5b8/packages/serialization/json/src/jsonSerializationWriter.ts#L139)
and [that](https://github.com/microsoft/kiota-serialization-json-dotnet/blob/295cfe1218b09d452e71443022d9c9ddcd7611c5/src/JsonSerializationWriter.cs#L289) (the later somehow missing from TypeScript???).
But yes, you can also go ahead with the generation PR.

salmannotkhan:
That's what I was thinking about lol. But it works find even without that. I tested enum array before renaming the method. Do I have to change anything else in this repo?

baywet:
Can you at least add an additional unit test that demonstrates symmetry please?

salmannotkhan:
Did you mean [this](https://github.com/microsoft/kiota-typescript/pull/752/files#diff-1f2a5d21439bdf7808a752b2ec5d2f85776d137d51cb7986a32f068a4056a669)?

salmannotkhan:
If you want me to add anything else, feel free to mention it

baywet:
This is a deserialization test. I mean a test where you'd serialize and deserialize and compare the results.

salmannotkhan:
Okay let me check

salmannotkhan:
I think you mean [this](https://github.com/microsoft/kiota-typescript/blob/main/packages/serialization/json/test/common/jsonSerializationWriter.ts) but we don't have any method for `writeCollectionOfEnumValues` in `jsonSerializationWriter` maybe because it's just a normal string array.

baywet:
yes, as far as I can remember they are treated as scalar values. Thanks for double checking

baywet (dismissed):

PR #772, opened 2023-08-02
- bumps node version

baywet:
14 is out of support, 20 is current, aligning https://nodejs.dev/en/about/releases/


sebastienlevert:
We're closing? @baywet

Issue #804, opened 2023-08-24
Serializing body with string property values in generated POST requests


jimmywim:
I have a client generated with Kiota from an OpenAPI file, and several POST methods will send up objects with string value properties, that have all sorts of characters such as newlines, quotes etc.

It seems these string values aren't serialized correctly when making a POST method.

Before kiota, we'd just JSON.stringify the payload and send it up, which would encode string value properties correctly, but it looks like Kiota is writing string value properties verbatim, which breaks the payload and it's no longer valid JSON.

Example:

Say we have an object that we're wanting to POST:

```
const myObj: SomeObjectType = {
  property1: true,
  property2: `This is a "complex" string type with some items on a \r\ndifferent line`
};

await client.api.someController.someAction.post(myObj);
```

Then request will fail as the double quotes and the new line character in that string invalidate the generated JSON.

I've noticed that jsonSerializationWriter does this:

```
    value && this.writer.push(`"${value}"`);
```
https://github.com/microsoft/kiota-typescript/blob/fb4c6848cb05c7d0b5f2aa642776f47103da4860/packages/serialization/json/src/jsonSerializationWriter.ts#L32C5-L32C45

whereas it probably should do this:

```
  value && this.writer.push(JSON.stringify(value));
```

baywet:
Thanks for reporting this! I believe you are correct in your analysis of the problem and in the suggested solution.
We could also take this opportunity to reduce duplication of code and map all the methods that "end up writing strings" to the writeStringValue method (thinking of Date, DateOnly, TimeOnly, Guid...)

Would you be willing to submit a pull request to correct this?

jimmywim:
Sure, can do.

PR #805, opened 2023-08-24
Fixed: complex string serialization

jimmywim:
Fixes: #804 

Uses native JSON stringify to serialize an input string to ensure all special characters are encoded/escaped, for example, line breaks, quotation marks, etc.

This allows the JSON serializer to create a valid JSON payload when a property has a complex string.

TODO:
Should we refactor, at this point, all `writeXValue` methods in this serializer to forward the call to the `writeStringValue`?

### DIFF
--- a/packages/serialization/json/src/jsonSerializationWriter.ts
+++ b/packages/serialization/json/src/jsonSerializationWriter.ts
@@ -53,8 +53,11 @@ export class JsonSerializationWriter implements SerializationWriter {
   public writeDateOnlyValue = (key?: string, value?: DateOnly): void => this.writeStringValue(key, value?.toString());
   public writeTimeOnlyValue = (key?: string, value?: TimeOnly): void => this.writeStringValue(key, value?.toString());
   public writeDurationValue = (key?: string, value?: Duration): void => this.writeStringValue(key, value?.toString());
-  public writeNullValue = (key?: string): void => this.writeStringValue(key, `null`);
-
+  public writeNullValue = (key?: string): void => {
+    key && this.writePropertyName(key);
+    this.writer.push(`null`);
+    key && this.writer.push(JsonSerializationWriter.propertySeparator);
+  };
   public writeCollectionOfPrimitiveValues = <T>(
     key?: string,
     values?: T[],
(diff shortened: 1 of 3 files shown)


baywet (changes requested):
Thanks for the contribution.
Could you please:
- add a unit test that covers the scenario to prevent regressions (yarn install --ignore-engines to get deps installed, yarn test to run the tests)
- (optional) map some of the WriteX methods (all the ones that write surrounding double quotes) to the writeStringValue one

packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -43,41 +43,18 @@ export class JsonSerializationWriter implements SerializationWriter {
...
-  public writeNullValue = (key?: string): void => {
-    key && this.writePropertyName(key);
-    this.writer.push(`null`);
-    key && this.writer.push(JsonSerializationWriter.propertySeparator);
-  };
+  public writeDateValue = (key?: string, value?: Date): void => this.writeStringValue(key, value?.toISOString());
+  public writeDateOnlyValue = (key?: string, value?: DateOnly): void => this.writeStringValue(key, value?.toString());
+  public writeTimeOnlyValue = (key?: string, value?: TimeOnly): void => this.writeStringValue(key, value?.toString());
+  public writeDurationValue = (key?: string, value?: Duration): void => this.writeStringValue(key, value?.toString());
+  public writeNullValue = (key?: string): void => this.writeStringValue(key, `null`);
```
baywet, line 56: I believe this one shouldn't have been changed since null doesn't need quotes in JSON. Other methods are fine.

packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -43,36 +43,16 @@ export class JsonSerializationWriter implements SerializationWriter {
       isValuePresent &&
       this.writer.push(JsonSerializationWriter.propertySeparator);
   };
-  public writeNumberValue = (key?: string, value?: number): void => {
-    key && value && this.writePropertyName(key);
-    value && this.writer.push(`${value}`);
-    key && value && this.writer.push(JsonSerializationWriter.propertySeparator);
-  };
+  public writeNumberValue = (key?: string, value?: number): void => this.writeStringValue(key, value?.toString());
```
baywet, line 46: sorry I missed that one, numbers don't need quotes, can you revert this one too please?

baywet (changes requested):
sorry, I missed one

packages/serialization/json/src/jsonSerializationWriter.ts
jimmywim: Done.

baywet (approved):
Thanks for making all the changes!

PR #819, opened 2023-08-30
Exclude popular editors in gitignore

rkodev:
Exclude popular editors in gitignore


baywet (changes requested):

baywet (approved):
Thanks for making the changes

PR #829, opened 2023-09-04
Switch to stduritemplate

### LINKED ISSUE #827: swap uri template for std uri template
@andreaTP as soon as you think you're ready, please go ahead and submit a PR here. One thing to be careful about for TypeScript is the type of module you're bundling to. For now we're supporting both commonjs and ESM with a target of es2017. If you have questions around those @koros can help with that.

andreaTP:
Fix #827

### DIFF
--- a/packages/abstractions/package.json
+++ b/packages/abstractions/package.json
@@ -31,15 +31,14 @@
   "homepage": "https://github.com/microsoft/kiota#readme",
   "devDependencies": {
     "@types/node": "^20.1.1",
-    "@types/uri-template-lite": "^19.12.1",
     "@types/uuid": "^9.0.0"
   },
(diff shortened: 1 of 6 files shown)


andreaTP:
The integration CI job with Kiota upstream is consistently failing:
https://github.com/microsoft/kiota-typescript/actions/runs/6074742576/job/16479447219?pr=830

The "normal CI" is affected by some instability of GH Actions:
 - failure: https://github.com/microsoft/kiota-typescript/actions/runs/6072440060/job/16477771098?pr=829
 - but the incident should be already over: https://www.githubstatus.com/incidents/76xp2jd3px64

koros (approved):

baywet (approved):

Issue #393, opened 2022-10-18
RequestBuilderGetQueryParameters sometimes have methods?


sebastienlevert:
When using the following kiota command to generate a client `kiota generate -l typescript -o nytmoviereviews -d https://raw.githubusercontent.com/nytimes/public_api_specs/master/movie_reviews/movie_reviews_v2.json -c NytMovieClient -n Nyt`, the `JsonRequestBuilderGetQueryParameters` has a method and becomes impossible to use without reimplementing the method in the call we are doing (when using partial objects in TypeScript).

![image](https://user-images.githubusercontent.com/7620955/196564158-a399cb12-fe73-40e4-83ce-0de238b14ee0.png)

```typescript

export class JsonRequestBuilderGetQueryParameters {
    /** Set this parameter to Y to limit the results to NYT Critics' Picks. To get only those movies that have not been highlighted by Times critics, specify critics-pick=N. (To get all reviews regardless of critics-pick status, simply omit this parameter.) */
    public criticsPick?: string | undefined;
    /** Positive integer, multiple of 20 */
    public offset?: number | undefined;
    /** Single date: YYYY-MM-DDStart and end date: YYYY-MM-DD;YYYY-MM-DDThe opening-date is the date the movie's opening date in the New York region. */
    public openingDate?: string | undefined;
    /** Sets the sort order of the results.Results ordered by-title are in ascending alphabetical order. Results ordered by one of the date parameters are in reverse chronological order.If you do not specify a sort order, the results will be ordered by publication-date. */
    public order?: string | undefined;
    /** Single date: YYYY-MM-DDStart and end date: YYYY-MM-DD;YYYY-MM-DDThe publication-date is the date the review was first published in The Times. */
    public publicationDate?: string | undefined;
    /** Search keywords; matches movie title and indexed termsTo limit your search to exact matches only, surround your search string with single quotation marks (e.g., query='28+days+later'). Otherwise, responses will include partial matches ("head words") as well as exact matches (e.g., president will match president, presidents and presidential).    If you specify multiple terms without quotation marks, they will be combined in an OR search.    If you omit the query parameter, your request will be equivalent to a reviews and NYT Critics' Picks request. */
    public query?: string | undefined;
    /** Include this parameter to limit your results to reviews by a specific critic. Reviewer names should be formatted like this: Manohla Dargis. */
    public reviewer?: string | undefined;
    /**
     * Maps the query parameters names to their encoded names for the URI template parsing.
[...]

PR #852, opened 2023-09-19
Adds request parameter name

rkodev:
Adds a map with the request parameter names that translates parameter names when provided.

Partially addresses https://github.com/microsoft/kiota-typescript/issues/393


baywet (changes requested):
@rkodev there are linting failures here, would you mind having a look before we review please?

PR #862, opened 2023-09-26
Restore backing store support

koros:
This PR contains changes to add the backing store support for kiota-typescript generated code


baywet (changes requested):

packages/abstractions/src/store/backingStoreUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+export const BackingStoreKey = "backingStore";
+
+/**
+   * Check if the object is an instance a BackedModel
+   * @param obj 
+   * @returns 
+   */
+export function isBackingStoreEnabled(fields: Record<string, (node: ParseNode) => void> ): boolean {
+    // Check if the fields contain the backing store key
+    return Object.keys(fields).includes(BackingStoreKey);
```
baywet, line 12: so you're effectively planning to add a "flag" property in the serialization/deserialization fields at generation when the backing store is enabled to know whether it's been enabled or not if I get this right? It's smart, let's make sure it doesn't have side effects on the developer experience.

packages/serialization/form/src/formParseNode.ts
```diff
@@ -78,7 +81,9 @@ export class FormParseNode implements ParseNode {
   public getObjectValue = <T extends Parsable>(
     parsableFactory: ParsableFactory<T>,
   ): T => {
-    const value: T = {} as T;
+    const temp: T = {} as T;
+    const enableBackingStore = isBackingStoreEnabled(parsableFactory(this)(temp));
+    const value: T = enableBackingStore ? new Proxy({}, createBackedModelProxyHandler<T>()) as T : temp;
```
baywet, line 86: could we just not pass "temp" into the proxy first argument as we create the proxy? so we avoid creating yet another object?

baywet (changes requested):

packages/abstractions/src/store/backingStoreUtils.ts
koros: Yeah that the plan, in this case if the backing store is enabled then the serialization/deserialization field will contain the backingStore property. 

Something like this:
``` typescript
/** Defines the contracts for a model that is backed by a store. */
export interface BackedModel {
  /**
   * Gets the store that is backing the model.
   */
  backingStore: BackingStore;
}

export function deserializeIntoBackedModel(
  backedModel: BackedModel | undefined = { backingStore: BackingStoreFactorySingleton.instance.createBackingStore() } // maybe make the backingStore field optional here
): Record<string, (node: ParseNode) => void> {
  return {
	"backingStore": n => { backedModel.backingStore = BackingStoreFactorySingleton.instance.createBackingStore(); },
  };
}


export function deserializeIntoAttachment(attachment: Attachment | undefined = {} as Attachment) : Record<string, (node: ParseNode) => void> {
    return {
	...deserializeIntoBackedModel(attachment),
        ...deserializeIntoEntity(attachment),
        "contentType": n => { attachment.contentType = n.getStringValue(); },
        "isInline": n => { attachment.isInline = n.getBooleanValue(); },
        "lastModifiedDateTime": n => { attachment.lastModifiedDateTime = n.getDateValue(); },
        "name": n => { attachment.name = n.getStringValue(); },
        "size": n => { attachment.size = n.getNumberValue(); },
    }
}
```

baywet: We can probably optimize further by replacing this `backedModel.backingStore = BackingStoreFactorySingleton.instance.createBackingStore();` (which might be costly depending on the backing store implementation) by a call to a method that creates a "dummy" (or uses just the in memory one) backing store in abstractions, thoughts?

baywet: maybe primitive (boolean) would be better instead now that you mention it

PR #878, opened 2023-10-04
Add a tryAdd method for request headers

andreaTP:
Preparation work for: https://github.com/microsoft/kiota/pull/3401


baywet:
@rkodev we're still seeing race conditions in the file grouping logic and the imports logic FYI

PR #921, opened 2023-10-30
- fixes a bug where encoding would be wrong for special characters

### LINKED ISSUE #920: Encoding Issue with Polish Characters in Request Payload
I am experiencing an encoding issue with Polish characters when sending requests. Before sending the request, I verified the value of my string variable by logging it to the console using console.log, and it displayed correctly as "Błonie". However, in the request payload, the variable value is incorrectly encoded as "BBonie". This issue persists with other words containing Polish characters. I am unable to determine the cause of this encoding error.

I have encountered this issue while using TypeScript Client Generator versions v1.8.0-preview.202310260001 and v1.7.0. Any assistance in resolving this encoding issue would be greatly appreciated.

baywet:
fixes #920

### DIFF
--- a/packages/abstractions/package.json
+++ b/packages/abstractions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/kiota-abstractions",
-  "version": "1.0.0-preview.29",
+  "version": "1.0.0-preview.30",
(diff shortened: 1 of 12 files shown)


illunix:
When this feature will be available, I need it urgent

andrueastman (approved):

PR #984, opened 2023-12-05
feature/models groups ts

### LINKED ISSUE #983: Header management should using a Record<String, String>
When appending headers to a request, we should be using a familiar approach. Currently, we need to specify a full `Headers` (from `kiota-abstractions`) like this:

```ts
headers: new Headers([['CopnsistencyLevel', new Set(['Eventual'])]]),
```

It should be closer to something like this : 

```ts
headers: { "ConsistencyLevel": "Eventual" }
```

baywet:
fixes #983
related https://github.com/microsoft/kiota/issues/3843


packages/abstractions/src/headers.ts
sebastienlevert, line 1: Having this class called "Headers" will be very challenging later. Knowing that there is a built-one in JS / Node that is also called Headers, I feel we'll get plenty of issues with this. I would suggest renaming it.

baywet: Do you have a better name to suggest? People can always alias the import should they run into a conflict. 
I can't see a scenario where they'd need to import it at the moment:
- passing headers to a request before executing it on the fluent API now support object notation.
- manipulating the request information (from the toXXXRequestInformation) they should only access the property and methods under it.
- using observability options, same rationale.

Issue #1017, opened 2024-01-09
remove string empty test when adding query parameters


baywet:
related to https://github.com/microsoft/kiota-abstractions-dotnet/issues/172

PR #1024, opened 2024-01-15
allow empty query parameters to be set.

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/1017

### DIFF
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -67,7 +67,7 @@ export class RequestInformation {
     } else {
       const data = {} as { [key: string]: unknown };
       for (const key in this.queryParameters) {
-        if (this.queryParameters[key]) {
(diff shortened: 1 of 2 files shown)


andrueastman (approved):

packages/abstractions/src/requestInformation.ts
```diff
@@ -67,7 +67,7 @@ export class RequestInformation {
     } else {
       const data = {} as { [key: string]: unknown };
       for (const key in this.queryParameters) {
-        if (this.queryParameters[key]) {
+        if (this.queryParameters[key] != null && this.queryParameters[key] !== undefined) {
```
baywet, line 70:
```suggestion
        if (this.queryParameters[key] !== null && this.queryParameters[key] !== undefined) {
```

baywet (changes requested):

baywet (approved):

Issue #467, opened 2022-12-15
RequestOptions implementations constructors should accept an interface instead of parameters


sebastienlevert:
This would allow developers to override only parts of the available options.

For example, instead of this:

```typescript
const option = new RetryHandlerOptions(3, 10, () => true);
```

I would love to set just the number of `maxRetries` with this:

```typescript
const option = new RetryHandlerOptions({ maxRetries: 10 });
```

PR #1033, opened 2024-01-22
RequestOptions implementations constructors should accept an interface instead of parameters

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/467

### DIFF
--- a/packages/http/fetch/src/middlewares/options/headersInspectionOptions.ts
+++ b/packages/http/fetch/src/middlewares/options/headersInspectionOptions.ts
@@ -8,7 +8,15 @@ export const HeadersInspectionOptionsKey = "HeadersInspectionOptionsKey";
  * @property {boolean} [inspectResponseHeaders = false] - Whether to inspect response headers
  */
 export interface HeadersInspectionOptionsParams {
+	/**
+	 * @member
+	 * whether to inspect request headers
+	 */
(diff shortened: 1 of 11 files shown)


packages/http/fetch/src/middlewares/options/headersInspectionOptions.ts
```diff
@@ -27,18 +48,35 @@ export class HeadersInspectionOptions implements RequestOption {
...
-	 * @param {boolean} [inspectRequestHeaders = false] - Whether to inspect request headers
-	 * @param {boolean} [inspectResponseHeaders = false] - Whether to inspect response headers
+	 * @param {HeadersInspectionOptionsParams} [options = {}] - The headers inspection options value
 	 * @returns An instance of HeadersInspectionOptions
 	 */
-	public constructor(
-		public inspectRequestHeaders = false,
-		public inspectResponseHeaders = false,
-	) {}
+	public constructor(options: HeadersInspectionOptionsParams = {} as HeadersInspectionOptionsParams) {
```
baywet, line 75: as far as I understand this will request people to either pass nothing, or the whole set of properties.
Could we type the argument with `Partial<HeadersInspectionOptionsParams>` instead?

baywet (changes requested):

Issue #985, opened 2023-12-05
The current middleware pipeline infrastructure should be aligned with the spec


sebastienlevert:
In the current Graph JS SDK, we can inject a middleware that will be able to inject some new capabilities (in my case, a custom proxy that could connect to the Dev Proxy to fake throttling, mock responses, etc.).

```ts
import { Middleware } from "@microsoft/microsoft-graph-client";
import { Context } from "@microsoft/microsoft-graph-client";
import { HttpsProxyAgent } from "hpagent";

export class ProxyMiddleware implements Middleware {
  private url: string;

  private nextMiddleware!: Middleware;

  public constructor(url: string = "http://localhost:8000") {
    this.url = url;
  }

  public async execute(context: Context): Promise<void> {
    if (context.options) {
      context.options.agent = new HttpsProxyAgent({
        proxy: this.url,
        rejectUnauthorized: false,
      });
    }

    return await this.nextMiddleware.execute(context);
  }

  public setNext(next: Middleware): void {
    this.nextMiddleware = next;
  }
}
```

I'm trying to achieve something similar with the current state of Kiota TypeScript and can't get it to work. What am I missing?

```ts
import { RequestOption } from '@microsoft/kiota-abstractions';
import { Middleware } from '@microsoft/kiota-http-fetchlibrary';
import { HttpsProxyAgent } from 'hpagent';

export class ProxyMiddleware implements Middleware {
  private url: string;
  public next: Middleware | undefined;

  public constructor(url: string = 'http://127.0.0.1:8000') {
    this.url = url;
  }

  public async execute(
    url: string,
    requestInit: RequestInit,
    requestOptions?: Record<string, RequestOption>,
  ): Promise<Response> {
    if (requestOptions) {
      requestOptions['agent'] = {
        agent: new HttpsProxyAgent({
          proxy: this.url,
          rejectUnauthorized: false,
        }),
        getKey: () => 'agent',
      } as RequestOption;
    }

    return (
      (await this.next?.execute(url, requestInit, requestOptions)) ??
      Promise.reject(new Error('The next middleware is not set.'))
    );
  }
}
```

baywet:
This is yet another area where we have divergences between typescript and other languages for some reason....

This doesn't exist in other languages

https://github.com/microsoft/kiota-typescript/blob/eb7c7aea28ea2c90efcdab26e0243aa0c58d8890/packages/http/fetch/src/middlewares/middlewareFactory.ts#L34

This is missing

[...]

PR #1043, opened 2024-01-25
bugfix/multiple fixes

baywet:
- - removes unused factory
- - adds missing export
- - doc comments fixups
- - typo fixes


packages/http/fetch/src/kiotaClientFactory.ts
```diff
@@ -1,16 +0,0 @@
-import type { Middleware } from "./middlewares/middleware";
```
andrueastman, line 1: To be addressed in https://github.com/microsoft/kiota-typescript/issues/985 I believe

andrueastman (approved):

PR #1069, opened 2024-02-13
fix malformed body and include additional optional data

### LINKED ISSUE #1068: MultiPart Request Error
The MultiPart body is malformed, the string value contains an additional **\r\n** and the server doesn't accept it when the values are validated.
Also, when uploading a file, "filename" is required by Django, but there is no way to provide additional keys and values to the message part.

p30arena:
Fixes #1068

### DIFF
--- a/packages/abstractions/src/multipartBody.ts
+++ b/packages/abstractions/src/multipartBody.ts
@@ -97,6 +97,7 @@ interface MultipartEntry {
   serializationCallback?: ModelSerializerFunction<Parsable>;
 }
 
+
 export function serializeMultipartBody(
   writer: SerializationWriter,
   multipartBody: Partial<MultipartBody> = new MultipartBody(),
@@ -123,16 +124,19 @@ export function serializeMultipartBody(
     if (first) {
       first = false;
     } else {
-      writer.writeStringValue(undefined, "");
+      writer.writeStringValue(undefined, "\r\n");
     }
     writer.writeStringValue(undefined, "--" + boundary);
+    writer.writeStringValue(undefined, "\r\n");
     const part = parts[partName];
     writer.writeStringValue("Content-Type", part.contentType);
+    writer.writeStringValue(undefined, "\r\n");
     writer.writeStringValue(
       "Content-Disposition",
       'form-data; name="' + part.originalName + '"',
     );
-    writer.writeStringValue(undefined, "");
+    writer.writeStringValue(undefined, "\r\n");
+    writer.writeStringValue(undefined, "\r\n");
     if (typeof part.content === "string") {
       writer.writeStringValue(undefined, part.content);
     } else if (part.content instanceof ArrayBuffer) {
@@ -169,8 +173,9 @@ export function serializeMultipartBody(
       );
     }
   }
-  writer.writeStringValue(undefined, "");
+  writer.writeStringValue(undefined, "\r\n");
   writer.writeStringValue(undefined, "--" + boundary + "--");
+  writer.writeStringValue(undefined, "\r\n");
 }
 
 export function deserializeIntoMultipartBody(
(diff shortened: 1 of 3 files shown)


p30arena:
@microsoft-github-policy-service agree company="Donusoft"

packages/abstractions/src/serialization/serializationWriter.ts
```diff
@@ -19,7 +19,7 @@ export interface SerializationWriter {
    * @param key the key to write the value with.
    * @param value the value to write to the stream.
    */
-  writeStringValue(key?: string | undefined, value?: string | undefined): void;
```
baywet, line 22: please revert this change, we want the serialization writer interfaces to be consistent across languages

packages/serialization/multipart/src/multipartSerializationWriter.ts
```diff
@@ -42,7 +42,9 @@ export class MultipartSerializationWriter implements SerializationWriter {
       }
       this.writeRawStringValue(value);
     }
-    this.writeRawStringValue("\r\n");
```
baywet, line 45: do you think removing this line instead, and adding `\r\n` to the calls to writeStringValue that require it would be a viable alternative here?

packages/abstractions/src/multipartBody.ts
```diff
@@ -128,13 +131,18 @@ export function serializeMultipartBody(
     writer.writeStringValue(undefined, "--" + boundary);
     const part = parts[partName];
     writer.writeStringValue("Content-Type", part.contentType);
+    let dispositionValue = 'form-data; name="' + part.originalName + '"';
```
baywet, line 134: can you add additional tests here https://github.com/microsoft/kiota-typescript/blob/main/packages/serialization/multipart/test/common/multipartSerializationWriter.ts for that additional behaviour please?

baywet (changes requested):
Thanks for the contribution! A couple of remarks to ensure this gets merged swiftly.

packages/abstractions/src/serialization/serializationWriter.ts
p30arena: I agree

packages/serialization/multipart/src/multipartSerializationWriter.ts
p30arena: can't we make writeRawStringValue public?
if so I could replace this line with:
writeRawStringValue

baywet: ideally we wouldn't. The idea is that the Serializer implementations only expose the methods defined in the interfaces.

packages/abstractions/src/multipartBody.ts
p30arena: addOrReplacePart changes are not consistent across languages, isn't it a problem?

p30arena: https://www.ietf.org/rfc/rfc2388.txt

it seems that filename is not mandatory.

> The original local file name may be supplied as well, either as a
>    "filename" parameter either of the "content-disposition: form-data"
>    header or, in the case of multiple files, in a "content-disposition:
>    file" header of the subpart. The sending application MAY supply a
>    file name; if the file name of the sender's operating system is not
>    in US-ASCII, the file name might be approximated, or encoded using
>    the method of RFC 2231.

baywet: > addOrReplacePart changes are not consistent across languages, isn't it a problem?

Can you provide more context for this please?

p30arena:
@baywet so what's the status?

baywet:
@p30arena sorry I've been out for medical reasons most of this week.
I'm waiting for two things:
- resolution of this conversation https://github.com/microsoft/kiota-typescript/pull/1069#discussion_r1487881255
- added unit tests to prevent regressions

shemogumbe:
Closing in favor of https://github.com/microsoft/kiota-typescript/pull/1423

PR #1084, opened 2024-02-27
[TypeScript] Untyped Nodes

andrueastman:
This PR is related to https://github.com/microsoft/kiota/issues/2319

It adds a number of types that implement the Parsable interface to enable the representation of types that are unknown at generation time to unblock the serialization and deserialization of properties of unknown types. 

It also adds functionality to the serialization library of Json to be able to serialize and deserialize the types as well as tests for the scenarios.

Generation of these types is unblocked in https://github.com/microsoft/kiota/pull/4095


koros (approved):

packages/abstractions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/kiota-abstractions",
-  "version": "1.0.0-preview.44",
```
baywet, line 3: I've already bumped to 45 this morning in #1100. We'll need to rebase and bump all libraries one more preview

packages/abstractions/src/serialization/untypedArray.ts
```diff
@@ -0,0 +1,19 @@
+import { isUntypedNode, UntypedNode } from "./untypedNode";
+
+export class UntypedArray extends UntypedNode {
```
baywet, line 3: I'm a bit affraid that javascript/typescript developers will hate us if we make them use classes to build JSON payloads (or other formats).
Do you think we could move to an interface approach instead?

baywet (changes requested):

packages/abstractions/src/serialization/untypedArray.ts
andrueastman: I hear you on this. 
Let me spike this one and propose an alternative so that we can look at to evaluate the two.

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -78,6 +93,39 @@ export class JsonParseNode implements ParseNode {
...
+      } else if (valueType === "string") {
+        value = createUntypedString(this._jsonNode as string) as any as T;
+      } else if (valueType === "number") {
+        value = createUntypedNumber(this._jsonNode as number) as any as T;
+      } else if (Array.isArray(this._jsonNode)) {
+        const nodes: UntypedNode[] = [];
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        (this._jsonNode as any[]).forEach((x) => {
+          nodes.push(
+            new JsonParseNode(x).getObjectValue(
```
baywet, line 110: is there anything that guarantees this is going to be an array of objects instead of an array of strings for example at this stage?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -78,6 +93,39 @@ export class JsonParseNode implements ParseNode {
...
+              createUntypedNodeFromDiscriminatorValue,
+            ),
+          );
+        });
+        value = createUntypedArray(nodes) as any as T;
+      } else if (this._jsonNode && valueType === "object") {
+        const properties: Record<string, UntypedNode> = {};
+        Object.entries(this._jsonNode as any).forEach(([k, v]) => {
+          properties[k] = new JsonParseNode(v).getObjectValue(
```
baywet, line 119: same remark, are we guaranteed properties will all be objects at this stage?

baywet (changes requested):
Thanks! Much better, I believe we still have a few things to finish up

packages/abstractions/src/serialization/untypedBoolean.ts
```diff
@@ -0,0 +1,17 @@
+import { UntypedNode } from "./untypedNode";
+
+export interface UntypedBoolean extends UntypedNode {
+  getValue(): boolean;
+}
+
+export function isUntypedBoolean(node: UntypedNode): node is UntypedBoolean {
```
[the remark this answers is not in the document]
andrueastman: Yeah, its a type guard syntax. 
When you define it like this, the compiler will know that the true branch of the code is of the `UntypedBoolean` type and therefore you can access members in a typed fashion without having to do any runtime checks. 
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/serialization/json/src/jsonParseNode.ts
andrueastman: Resolving this to address in other comment.

packages/serialization/json/src/jsonParseNode.ts
andrueastman: That is true. The array is not necessarily an array of objects. 
However, calling `getObjectValue` with the factory as `createUntypedNodeFromDiscriminatorValue` will assure that the type check made by `isUntypedNode`(line 96) on the next recursive call will be true. 
Therefore the array/object will be populated by an UntypedNode type e.g. it will be an array of `UntypedString` rather than `UntypedObject`

packages/abstractions/src/serialization/untypedArray.ts
```diff
@@ -0,0 +1,21 @@
+import { isUntypedNode, UntypedNode } from "./untypedNode";
+
+export interface UntypedArray extends UntypedNode {
```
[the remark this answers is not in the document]
andrueastman: Resolved in latest updates.

packages/abstractions/src/serialization/untypedBoolean.ts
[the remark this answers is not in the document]
baywet: everyday I learn a little more! Thanks for sharing!

baywet (approved):
Thanks for making the changes!

Issue #1083, opened 2024-02-26
Adopt built-in fetch


bterlson:
The current approach of using `node-fetch` is less than ideal because it introduces a fair bit of friction in real projects (especially module related 😁) and is big and slow, and Node has built-in fetch in Node 18+ that works a lot better. It also allows sharing more code between browser and node, which is a win.

sebastienlevert:
Now that we have bumped out Node version to at least 18 IIRC (maybe even 20 @baywet) I think we could consider this. This would help with bundle size and standardization of the API.

Any objection on the topic @baywet or @koros?

baywet:
Node18 still has [security support until April 2025](https://endoflife.date/nodejs), so I think targeting node 18/20 for the GA makes sense.
[The fetch API in node is only stable with 21](https://blog.logrocket.com/fetch-api-node-js/) but has been [enabled by default since node18](https://nodejs.org/api/globals.html#fetch) and the API surface from node18 is all I believe we need.

baywet:
Additionally we are still on node-fetch 2.X because it's the last version that supports commonjs modules. See discussions ongoing in #1055.
Removing that dependency will make it easier to both:
- keep supporting commonjs
- drop support for commonjs

PR #1085, opened 2024-02-27
Align middleware with spec

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/985


packages/serialization/form/src/browser/formParseNodeFactory.ts
```diff
@@ -13,7 +13,7 @@ export class FormParseNodeFactory implements ParseNodeFactory {
     if (!content) {
       throw new Error("content cannot be undefined of empty");
     } else if (!contentType) {
-      throw new Error("content type cannot be undefined or empty");
+      throw new Error("content type cannot be undefined or pempty");
```
andrueastman, line 16:
```suggestion
      throw new Error("content type cannot be undefined or empty");
```

packages/http/fetch/test/node/httpClient.ts
```diff
@@ -9,6 +9,7 @@ import { assert } from "chai";
 
 import { CustomFetchHandler, HeadersInspectionHandler, HttpClient, ParametersNameDecodingHandler, RedirectHandler, RetryHandler, UserAgentHandler } from "../../src";
 import { DummyFetchHandler } from "../common/middleware/dummyFetchHandler";
+import { validateHeaderName } from "http";
```
andrueastman, line 12:
```suggestion
```

packages/http/fetch/src/middlewares/browser/middlewareFactory.ts
```diff
@@ -27,10 +16,14 @@ export class MiddlewareFactory {
...
 	 * @returns an array of the middleware handlers of the default middleware chain
 	 */
-	public static getDefaultMiddlewareChain(customFetch: (request: string, init: RequestInit) => Promise<Response> = fetch as any): Middleware[] {
+	public static getDefaultMiddlewares(customFetch: (request: string, init: RequestInit) => Promise<Response> = fetch as any): Middleware[] {
 		// Browsers handles redirection automatically and do not require the redirectionHandler
-
-		const middlewareArray: Middleware[] = [new RetryHandler(), new ParametersNameDecodingHandler(), new UserAgentHandler(), new HeadersInspectionHandler(), new CustomFetchHandler(customFetch)];
-		return middlewareArray;
+		return [
+            new RetryHandler(),
```
andrueastman, line 22: Just to confirm here, the reason we don't have the `RedirectHandler` in the browser config is because the browser will do that for us automatically. Yes?

koros: Yes, the browser should handle the redirects

baywet (changes requested):

packages/http/fetch/src/kiotaClientFactory.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Class containing function(s) related to the middleware pipelines.
+ */
+export class KiotaClientFactory {
+	/**
+	 * @public
+	 * @static
+	 * Returns an instance of HttpClient with the provided middlewares and custom fetch implementation
+	 * @returns a HttpClient instance
+	 */
+	public static create(customFetch: (request: string, init: RequestInit) => Promise<Response> = fetch as any, middlewares?: Middleware[]): HttpClient {
```
koros, line 15: @baywet instead of the optional parameters would it be helpful if I create additional signatures or a builder, example:
```javascript
create()
createClientWithMidlewares(middleware: Middleware[]);
createClientWithCustomFetch(customFetch: (request: string, init: RequestInit) => Promise<Response>);
createClientWithCustomFetchAndMiddleware(customFetch: (request: string, init: RequestInit) => Promise<Response>, middlewares: Middleware[])
```

baywet: I think this suggestion doesn't look like TypeScript usual approach: people expect optional parameters given there's good doc comments.
Now, maybe your suggestion is to move the middleware array to the first position? That I think would make more sense because I suspect most people will want to change only the middleware set rather than the fetch implementation under the hood.
Actually because of #1083, we might remove the custom fetch argument all together, but this will come in a latter PR I suspect.

sebastienlevert: +1 to @baywet comment!

Issue #1072, opened 2024-02-15
Polymorphism of buffer usage


baywet:
The libraries here make use of Buffer.from multiple places
https://github.com/microsoft/kiota-typescript/blob/dfaf4045855f358a44f1fa91b5b85da901b5d24e/packages/authentication/azure/src/azureIdentityAccessTokenProvider.ts#L85
https://github.com/microsoft/kiota-typescript/blob/dfaf4045855f358a44f1fa91b5b85da901b5d24e/packages/serialization/json/src/jsonParseNode.ts#L59
https://github.com/microsoft/kiota-typescript/blob/dfaf4045855f358a44f1fa91b5b85da901b5d24e/packages/serialization/json/src/jsonSerializationWriter.ts#L19
https://github.com/microsoft/kiota-typescript/blob/dfaf4045855f358a44f1fa91b5b85da901b5d24e/packages/serialization/text/src/textParseNode.ts#L31
https://github.com/microsoft/kiota-typescript/blob/dfaf4045855f358a44f1fa91b5b85da901b5d24e/packages/serialization/text/src/textSerializationWriter.ts#L20

This API is NOT AVAILABLE on browsers, we're using it to perform base64 encoding/decoding.

We should instead:
- define two base64 encode/decode methods in abstractions
- in the implementations test whether Buffer.from is available, if so, use that.
- in the implementations test whether atob/btoa are available, if so, use that.
- if none of those APIs are available, throw
- replace the calls to Buffer.from by the corresponding method.

musale:
Adding this to fixes https://github.com/microsoft/kiota-typescript/blob/cd0b0ba50d70e92df22ee1ffe204d95306156073/packages/abstractions/test/common/requestInformation.ts#L9 as url is also a nodejs API in the test usage and fails browser tests.

sebastienlevert:
@musale are you still working on this and still have time? This is blocking for public preview and we might need to re-assign to make progress. Please let us know!

petrhollayms:
Yep, confirmed with @musale today

musale:
@sebastienlevert @petrhollayms almost done with this. Fixing the conflicts and final touches based on https://github.com/microsoft/kiota-typescript/blob/main/packages/http/fetch/docs/design/isomorphic.md as I move over to ESM. I'll have an update on Monday.

PR #1099, opened 2024-03-08
chore: added vitest for node and browser testing in abstrations

musale:
This is a small change that is part of a bigger goal to improve our confidence in browser testing and update the codebase to adapt to node/browser environment without throwing errors:
- adds vitest as the test framework that has browser testing
- drops karma. eventually drop add-ins like sinon used for mocking and completely build on vitest.
- updates test imports to use vitest api
- adds browser testing for the package's `abstractions` and `authentication/azure`
- adds utility function to check the environment(node/browser) the code is running in
- Fixes browser incompatible code in `authentication.azure` package (part of #1072). Should update for the rest of the other instances.

### Status
The work in this pr is only done on 2 packages, abstractions and authentication/azure. This should give a basis and open discussion of what the end result for the goals above should look like.

- [x] `abstractions` package is updated to vitest and has browser tests
- [x] `authentication/azure` package is updated to vitest and has a trigger for browser tests that pass
- [ ] `authentication/spfx` is not updated
- [ ] `http` is not updated.
- [ ] `serialization/*` is not updated.


packages/authentication/azure/package.json
```diff
@@ -36,5 +38,11 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "@vitest/coverage-v8": "^1.3.1",
```
baywet, line 43: couldn't we simply make the dependencies workspace global instead of local to the package? I'm interested in reducing the number of dependabot PRs we'll get over time.

baywet (approved):
this long overdue and great work! thanks for taking care of it!

musale:
closing in favour of https://github.com/microsoft/kiota-typescript/pull/1102

PR #1102, opened 2024-03-11
chore: change to vitest and remove cjs bundling.

### LINKED ISSUE #1055: evaluate whether we still need commonjs support
Since we originally stood up this library the ecosystem has moved a lot:
- nodejs supports ESM, and all supported versions have them enabled by default now
- major community projects have moved to ESM only, including some in our build infrastructure (see #1041 where we can't upgrade to chai 5 because we still have commonjs based tests)
- we've established that the sub modules we have (fluent API, sub-namespaces for models...) can only be consumed in ESM (when the generated client is published as its own package) when working on msgraph-sdk-typescript.
- we're still on the 2.X branch of node-fetch because 3.X [moved to ESM](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md)

IMHO we should remove commonjs support after testing our main scenarios (SPFx, TeamsFX....) and BEFORE the GA so people don't end up in confusing variants + configuration misalignment issues.

### LINKED ISSUE #1072: Polymorphism of buffer usage

musale:
closes #1072, closes #1055

This is a change that is part of a bigger goal to improve our confidence in browser testing and update the codebase to adapt to node/browser environment without throwing errors:
- adds Vitest as the test framework that has browser testing.
- drops karma. eventually drop add-ins like Sinon used for mocking and completely build on Vitest.
- updates test imports to use Vitest API.
- adds browser testing for the packages.
- adds utility function to check the environment(node/browser) the code is running in.
- Fixes browser incompatible code in the codebase.
- Removes common js bundling tooling and support.

### Status

- [x] `abstractions` package is updated to vitest and has browser tests
- [x] `authentication/azure` package is updated to vitest and has a trigger for browser tests that pass
- [x] `authentication/spfx` is updated
- [x] `http` is updated.
- [x] `serialization/*` is updated.


packages/abstractions/test/common/authentication/validateProtocolTest.ts
```diff
@@ -1,10 +1,11 @@
 import { validateProtocol, isLocalhostUrl } from '../../../src/authentication';
-import { expect } from "chai";
+import { describe, it, expect } from "vitest";
 
 describe('validateProtocol', () => {
-    it('should throw an error for non-https and non-localhost URLs', () => {
-        expect(() => validateProtocol('http://example.com')).to.throw('Authentication scheme can only be used with https requests');
-    });
+    // TODO: fix this test
```
baywet, line 5: adding this comment as a reminder

baywet (approved):
great work so far! can't wait for this to be finalized

musale:
@baywet I'm having a blocker with completing this. When running the steps to build and test, there's an error with building that comes from having build instructions in tsconfig that outputs commonjs. I think it would be a good idea to make a decision on #1055 to unblock this. What do you think?

baywet:
Agreed, let's get confirmation on the PM side.
If Seb doesn't reply within a few days, please reach out to him directly

packages/abstractions/src/apiError.ts
```diff
@@ -1,16 +1,21 @@
-/** Parent interface for errors thrown by the client when receiving failed responses to its requests. */
```
baywet, line 1: for my education: are doc comment ignored for interfaces?

musale: I don't understand the context of your question @baywet

baywet: I saw that you removed the doc comment, can you explain why?

.github/workflows/build_test_validate.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     build:
         runs-on: ubuntu-latest
         environment:
-          name: build_test
+          name: Build and test
```
baywet, line 13: revert

packages/abstractions/src/apiError.ts
musale: Oh, I see! I didn't remove that intentionally; it must've slipped out during one of the merges. I'll revert it too.

koros:
Hi @musale,
Can you please give instructions on how to run the build and tests locally. this is what I get:

### `npm run build`
npm-run-build

### `npm run test`
npm-run-test

musale:
> Hi @musale, Can you please give instructions on how to run the build and tests locally. this is what I get:
> 
> ### `npm run build`
> npm-run-build
> 
> ### `npm run test`
> npm-run-test

Hi @koros this CI/CD pipeline and build is adapted for a Linux environment. I see you're using PowerShell and this will fail for commands like `clean` and path resolutions. Please use WSL or a Linux environment to test this.

packages/authentication/azure/package.json
```diff
@@ -30,11 +31,12 @@
   "homepage": "https://github.com/microsoft/kiota-typescript#readme",
   "dependencies": {
     "@azure/core-auth": "^1.5.0",
-    "@microsoft/kiota-abstractions": "^1.0.0-preview.50",
+    "@microsoft/kiota-abstractions": "*",
```
baywet, line 34: what's the rationale behind switching to a star reference?

packages/abstractions/src/store/backingStore.ts
```diff
@@ -1,55 +1,55 @@
 /**
- * Stores model information in a different location than the object properties. Implementations can provide dirty tracking capabilities, caching capabilities or integration with 3rd party stores.
+ * -------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.
+ * See License in the project root for license information.
```
baywet, line 4: still missing the doc comment here (and a couple of other places)

baywet (changes requested):
Getting there, still a couple of remarks

packages/authentication/azure/package.json
musale: Since we are using Lerna to manage the monorepo, in local development using the * allows using the latest changes in a package.

musale: Yes, the published version will use the latest version of abstractions. Is that going to be problematic?

baywet: no, that's what I expected. Thanks for confirming. Resolving

baywet (approved):
@musale there are new conflicts to resolve, but once that is done, I think we're good to merge here. Thanks for the great work!

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -49,11 +52,14 @@ export class JsonParseNode implements ParseNode {
 	public getByteArrayValue(): ArrayBuffer | undefined {
 		const strValue = this.getStringValue();
 		if (strValue && strValue.length > 0) {
-			return inNodeEnv() ? Buffer.from(strValue, "base64").buffer : new TextEncoder().encode(strValue);
+			return Buffer.from(strValue, "base64").buffer;
```
baywet, line 55: probably a regression during the merge here

musale: Ah yeah, good catch!

koros:
> > Hi @musale, Can you please give instructions on how to run the build and tests locally. this is what I get:
> > ### `npm run build`
> > npm-run-build
> > ### `npm run test`
> > npm-run-test
> 
> Hi @koros this CI/CD pipeline and build is adapted for a Linux environment. I see you're using PowerShell and this will fail for commands like `clean` and `tsconfig` file path resolutions. Please use WSL or a Linux environment to test this.

@musale this seems to resolve the issue on Win OS
``` JSON
    "build:@microsoft/*": "lerna run build --scope \"@microsoft/*\"",
    "clean": "lerna run --parallel --stream --scope \"@microsoft/*\" clean",
    "test:browser": "lerna run --scope \"@microsoft/*\" test:browser",
    "test:node": "lerna run --scope \"@microsoft/*\" test:node",
```

packages/test/package.json
```diff
@@ -5,16 +5,16 @@
...
-    "@microsoft/kiota-http-fetchlibrary": "^1.0.0-preview.22",
-    "@microsoft/kiota-serialization-form": "^1.0.0-preview.12",
-    "@microsoft/kiota-serialization-json": "^1.0.0-preview.21",
-    "@microsoft/kiota-serialization-text": "^1.0.0-preview.20"
+    "@microsoft/kiota-abstractions": "*",
+    "@microsoft/kiota-authentication-azure": "*",
+    "@microsoft/kiota-http-fetchlibrary": "*",
+    "@microsoft/kiota-serialization-form": "*",
+    "@microsoft/kiota-serialization-json": "*",
+    "@microsoft/kiota-serialization-text": "*"
```
koros, line 17:
```suggestion
    "@microsoft/kiota-serialization-text": "*",
    "@microsoft/kiota-serialization-multipart": "*"
```

packages/test/package.json
```diff
@@ -5,16 +5,16 @@
   "main": "index.js",
   "private": true,
   "scripts": {
-    "test:integrated": "mocha  --exit dist/cjs/tests/*.js"
+    "test:integrated": "vitest run"
```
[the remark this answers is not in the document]
koros: We should also probably change the `name` to `@microsoft/test-kiota-typescript-libraries`.
at the moment the whole package gets skipped when running build and tests because name doesn't match the filter criteria `"@microsoft/*"`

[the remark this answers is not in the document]
musale: @koros the `test` package is not a Microsoft namespace package, it's only used to run integrated tests. Running it as part of the Lerna workspace commands also introduces unnecessary commands like `build` and `clean`. It also has dependency issues needing IDs and secrets which, during Lerna runtime, we'll have to pass them on the test commands and it's not a recommended practice. As it is, running it as its own step in the CI/CD pipeline is sufficient.

baywet (approved):

koros (approved):

Issue #1113, opened 2024-03-19
Guid uri template path/query parameter values should be normalized to string


baywet:
the std uri template library does not handle guids, and wont because it's not a standard API across languages and they don't want to bring on additional dependencies.
We should normalize the value to string (no curlies) before we pass it to the uri template engine.
Originally reported:
https://github.com/std-uritemplate/std-uritemplate/issues/140
https://github.com/microsoft/kiota/issues/4335

illunix:
Can you help me @baywet ?

illunix:
Here?
https://github.com/microsoft/kiota-typescript/blob/cd45905d68f98ff7e2b7cf92ced4d3aacc838f4a/packages/abstractions/src/requestInformation.ts#L235-L259

baywet:
If we look [at the dotnet implementation](https://github.com/microsoft/kiota-abstractions-dotnet/blob/5a93d126e9ce30ae332e65f30df7b941e7d7c265/src/RequestInformation.cs#L90) we can see that the URI getter calls into an [additional method that's missing in typescript](https://github.com/microsoft/kiota-abstractions-dotnet/blob/5a93d126e9ce30ae332e65f30df7b941e7d7c265/src/RequestInformation.cs#L112) and which is responsible for handling enum query/path parameters as other scalar but non trivial types (guid, etc...)

This is effectively what we need to replicate [int typescript](https://github.com/microsoft/kiota-typescript/blob/cd45905d68f98ff7e2b7cf92ced4d3aacc838f4a/packages/abstractions/src/requestInformation.ts#L77)

You can then add unit tests [here](https://github.com/microsoft/kiota-typescript/blob/main/packages/abstractions/test/common/requestInformation.ts) to prevent any future regression.

Thank you for starting the pull request already!

illunix:
It seems hard for me and I'm not sure if I will be able to complete this but I'll try

illunix:
Also I don't understand this part with replication @baywet

PR #1116, opened 2024-03-19
Add GUID support for route params

illunix:
fixes https://github.com/microsoft/kiota-typescript/issues/1113

### DIFF
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -1,3 +1,4 @@
+import { Guid } from 'guid-typescript';
 import { trace } from "@opentelemetry/api";
 import { StdUriTemplate } from "@std-uritemplate/std-uritemplate";
 
@@ -252,6 +253,8 @@ export class RequestInformation implements RequestInformationSetContent {
               writer.writeTimeOnlyValue(undefined, value as any as TimeOnly);
             } else if (value instanceof Duration) {
               writer.writeDurationValue(undefined, value as any as Duration);
+            } else if (value instanceof Guid) {
+              writer.writeStringValue(undefined, value as any as string);
             } else if (valueType === "number") {
               writer.writeNumberValue(undefined, value as any as number);
             } else if (Array.isArray(value)) {
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -254,7 +254,7 @@ export class RequestInformation implements RequestInformationSetContent {
             } else if (value instanceof Duration) {
               writer.writeDurationValue(undefined, value as any as Duration);
             } else if (value instanceof Guid) {
-              writer.writeStringValue(undefined, value as any as string);
+              writer.writeGuidValue(undefined, value as any as Guid);
             } else if (valueType === "number") {
               writer.writeNumberValue(undefined, value as any as number);
             } else if (Array.isArray(value)) {


packages/abstractions/src/requestInformation.ts
```diff
@@ -252,6 +253,8 @@ export class RequestInformation implements RequestInformationSetContent {
               writer.writeTimeOnlyValue(undefined, value as any as TimeOnly);
             } else if (value instanceof Duration) {
               writer.writeDurationValue(undefined, value as any as Duration);
+            } else if (value instanceof Guid) {
```
baywet, line 256: This if statement is for the request body, it's good to fix the issue here as well as people will eventually run into it, but it's not going to address the initial issue.

packages/abstractions/src/requestInformation.ts
```diff
@@ -252,6 +253,8 @@ export class RequestInformation implements RequestInformationSetContent {
...
+              writer.writeStringValue(undefined, value as any as string);
```
baywet, line 257:
```suggestion
              writer.writeGuidValue(undefined, value as any as Guid);
```

baywet (changes requested):
Thanks for the contribution.
It needs further changes and unit tests.
I'll reply shortly to the questions on the issue to provide guidance

packages/abstractions/src/requestInformation.ts
yasaichi: @illunix 
I think there are 2 possible ways to convert a Guild value in path parameters to a string:

# Inside `RequestInformation.URL`

```typescript
public get URL(): string {
  const rawUrl = this.pathParameters[
    RequestInformation.raw_url_key
  ] as string;
  if (this.uri) {
    ...
  } else {
    const data = {} as { [key: string]: unknown };
    for (const key in this.queryParameters) {
      if (this.queryParameters[key] !== null && this.queryParameters[key] !== undefined) {
        data[key] = this.queryParameters[key];
      }
    }
    for (const key in this.pathParameters) {
      if (this.pathParameters[key]) {
        data[key] = this.pathParameters[key]; // <- here
      }
    }
    return StdUriTemplate.expand(this.urlTemplate, data);
  }
}

```

# Inside `apiClientProxifier`

```ts
export function apiClientProxifier<T extends object>(
  requestAdapter: RequestAdapter,
  pathParameters: Record<string, unknown>,
  navigationMetadata?: Record<string, NavigationMetadata>,
  requestsMetadata?: RequestsMetadata,
): T {
  if (!requestAdapter) throw new Error("requestAdapter cannot be undefined");
  if (!pathParameters) throw new Error("pathParameters cannot be undefined");
  return new Proxy({} as T, {
    get(target, property) {
      ...
      if (navigationMetadata) {
        const navigationCandidate = navigationMetadata[name];
        if (navigationCandidate) {
          ...
          return (...argArray: any[]) => {
            // navigation method like indexers or multiple path parameters
            const downWardPathParameters = getPathParameters(pathParameters);
            if (
              navigationCandidate.pathParametersMappings &&
              navigationCandidate.pathParametersMappings.length > 0
            ) {
              for (let i = 0; i < argArray.length; i++) {
                const element = argArray[i];
                downWardPathParameters[
                  navigationCandidate.pathParametersMappings[i]
                ] = element; // <- here
              }
            }
            return apiClientProxifier(
              requestAdapter,
              downWardPathParameters,
              navigationCandidate.navigationMetadata,
              navigationCandidate.requestsMetadata,
            );
          };
        }
        throw new Error(
          `couldn't find navigation property ${name} data: ${JSON.stringify(
            navigationMetadata,
          )}`,
        );
      }
    },
  });
}
```

I hope that this will help you. Thanks

Issue #1042, opened 2024-01-25
Add support for Enums or collection of enum values as root of the payload


baywet:
A good example of that is beta `POST /deviceManagement/chromeOSOnboardingSettings/disconnect` which returns `chromeOSOnboardingStatus` (enum)

Here are the changes we need to make:
1. Request adapter interface need to accept objects as well for the responseType argument of the "sendPrimitiveAsync" and "sendCollectionOfPrimitiveAsync" methods
2. reflect the changes to the implementation
3. add a case in the implementation to call getEnumValue/getCollectionOfEnumValues when an object is provided.
4. update kiota generation to use that method for enums, add a using to the Object of the enum, and pass that object (in CodeConstantWriter for the requests metadata constants)

+ required tests

andrueastman:
> 1. Request adapter interface need to accept objects as well for the responseType argument of the "sendPrimitiveAsync" and "sendCollectionOfPrimitiveAsync" methods

Just to confirm here that we should use "sendPrimitiveAsync" and "sendCollectionOfPrimitiveAsync" or align with other langs to add `sendEnumAsync` and `sendCollectionOfEnumsAsync`. 

cc @baywet

baywet:
I think it'd make more sense to align with the other languages to be specific about what's expected from the implementations.
Especially with the reasoning you've provided here https://github.com/microsoft/kiota-abstractions-dotnet/issues/199

PR #1117, opened 2024-03-20
Feature/support enum as root payload

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/1042

### DIFF
--- a/packages/http/fetch/test/common/fetchRequestAdapter.ts
+++ b/packages/http/fetch/test/common/fetchRequestAdapter.ts
@@ -147,14 +147,14 @@ describe("FetchRequestAdapter.ts", () => {
 					response.headers.set("Content-Type", "text/plain");
 					return Promise.resolve(response);
 				};
-				const mockFactory = new TextParseNodeFactory();//new JsonParseNodeFactory();
+				const mockFactory = new TextParseNodeFactory();
 				const requestAdapter = new FetchRequestAdapter(new AnonymousAuthenticationProvider(), mockFactory, undefined, mockHttpClient);
 				const requestInformation = new RequestInformation();
 				requestInformation.URL = "https://www.example.com";
 				requestInformation.httpMethod = HttpMethod.GET;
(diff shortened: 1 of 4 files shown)


packages/http/fetch/test/common/fetchRequestAdapter.ts
```diff
@@ -124,6 +135,55 @@ describe("FetchRequestAdapter.ts", () => {
...
+					return Promise.resolve(response);
+				};
+				const mockFactory = new TextParseNodeFactory();//new JsonParseNodeFactory();
+				const requestAdapter = new FetchRequestAdapter(new AnonymousAuthenticationProvider(), mockFactory, undefined, mockHttpClient);
+				const requestInformation = new RequestInformation();
+				requestInformation.URL = "https://www.example.com";
+				requestInformation.httpMethod = HttpMethod.GET;
+				const result = await requestAdapter.sendEnum(requestInformation, TestEnumObject, undefined);
+				assert.isDefined(result);
+				assert.equal(result, TestEnumObject.A);
```
baywet, line 157: shouldn't we be able to use test enum instead of test enum object?

koros: The nodeParser at the moment expects the object itself:

``` Typescript
public getEnumValue = <T>(type: any): T | undefined => {
    const rawValue = this.getStringValue();
    if (!rawValue) {
      return undefined;
    }
    return type[toFirstCharacterUpper(rawValue)];
  };
```

I couldn't find a way to verify if a node value is of desired `enum` type without any reference to the object

baywet: I get the fact that we need to pass the object to the request adapter method.
My comment was more about the assert line (157). I don't think we should be asking people to use the object, that's a poor developer experience. So I'm asking if we can update the unit test to work with the enum instead (for the assert part)

baywet (approved):

andrueastman (approved):

Issue #1079, opened 2024-02-16
Is it possible to skip undefined obj from json body? [Typescript]


illunix:
Hey, is it possible to skip undefined obj from json body?

Because right now when I do for example

` const cmd: SignUpCommand = {
      email: this.f.email.value!,
      business: undefined
    };`

it adds `business: {}` to json body but I dont wan't

baywet:
Thanks for reporting this @illunix 
I have transferred your issue to the relevant repository.
If we compare [the dotnet implementation](https://github.com/microsoft/kiota-serialization-json-dotnet/blob/7492e10c9f29cd63c4c3a6115698a9db1fc71896/src/JsonSerializationWriter.cs#L337) with the typescript one, we can see a divergence that would explain this behaviour.
https://github.com/microsoft/kiota-serialization-json-dotnet/blob/7492e10c9f29cd63c4c3a6115698a9db1fc71896/src/JsonSerializationWriter.cs#L337

(see the guard if statement)

Is this something you'd be willing to submit a pull request to address?

illunix:
I don't find myself in this code so rather not @petrhollayms

koros:
Re-opening the issue to ensure edge cases are covered as well since currently its using falsy evaluation to determine which values to skip.

koros:
Fixed by https://github.com/microsoft/kiota-typescript/pull/1133

PR #1120, opened 2024-03-21
add test case to ensure `undefined` fields are skipped while serializing.

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/1079

### DIFF
--- a/packages/serialization/json/test/common/jsonSerializationWriter.ts
+++ b/packages/serialization/json/test/common/jsonSerializationWriter.ts
@@ -78,6 +78,23 @@ describe("JsonParseNode", () => {
     const result = JSON.parse(contentAsStr);
     assert.equal(result.testComplexString, "Błonie");
   });
+  it("skip undefined objects from json body", async () => {
+    const inputObject: TestParser = {
+      testCollection: undefined,
+      testString: "test",
+      testObject: undefined,
+    };
+    const writer = new JsonSerializationWriter();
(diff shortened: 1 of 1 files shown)


packages/serialization/json/test/common/jsonSerializationWriter.ts
```diff
@@ -78,6 +78,23 @@ describe("JsonParseNode", () => {
...
+      testString: "test",
+      testObject: undefined,
+    };
+    const writer = new JsonSerializationWriter();
+    writer.writeObjectValue("", inputObject, serializeTestParser);
+    const serializedContent = writer.getSerializedContent();
+    const decoder = new TextDecoder();
+    const contentAsStr = decoder.decode(serializedContent);
+    const result = JSON.parse(contentAsStr);
+    console.log(JSON.stringify(result, null, 2));
```
baywet, line 93:
```suggestion
```

packages/serialization/json/test/common/jsonSerializationWriter.ts
```diff
@@ -78,6 +78,23 @@ describe("JsonParseNode", () => {
     const result = JSON.parse(contentAsStr);
```
baywet, line 78: we probably also need a unit test with the backing store and to correct the implementation divergence with dotnet.
https://github.com/microsoft/kiota-typescript/issues/1079#issuecomment-1959515442

baywet (changes requested):

andrueastman (approved):

Issue #1098, opened 2024-03-08
v1.12.0: Failed to execute 'fetch' on 'Window': Illegal invocation


seniorquico:
We upgraded our web-based, Vite/React SPA to the new Kiota v1.12.0 and upgraded our generated API client. The dependencies in our project were updated to include (as dictated by `kiota info ...`):

```
@microsoft/kiota-abstractions: 1.0.0-preview.44
@microsoft/kiota-http-fetchlibrary: 1.0.0-preview.43
@microsoft/kiota-serialization-json: 1.0.0-preview.44
```

Running the updated project, API requests fail with the following message and stack:

```
TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
    at CustomFetchHandler.<anonymous> (http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:598:29)
    at Generator.next (<anonymous>)
    at http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:185:67
    at new Promise (<anonymous>)
    at Object.__awaiter (http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:167:10)
    at CustomFetchHandler.execute (http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:597:24)
    at HeadersInspectionHandler.<anonymous> (http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:1600:44)
    at Generator.next (<anonymous>)
    at http://localhost:3000/node_modules/.vite/deps/@microsoft_kiota-http-fetchlibrary.js?v=8cce6e32:185:67
    at new Promise (<anonymous>)
```

Our client initialization code has been:

```ts
const myApiClient = createMyApiClient(new FetchRequestAdapter(new AnonymousAuthenticationProvider()))
```

Modifying the client initialization to use an explicitly bound `window.fetch` function "fixes" the error and gets us back to a working integration:

```ts
const myApiClient = createMyApiClient(new FetchRequestAdapter(new AnonymousAuthenticationProvider(), undefined, undefined, KiotaClientFactory.create(window.fetch.bind(window))))
```

[...]

PR #1127, opened 2024-03-27
fix fetch bug

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/1098


andrueastman (commented):
Could we do a version bump as well?

andrueastman (approved):

PR #1133, opened 2024-04-03
skip null/undefined fields when serializing

koros:
Fixes: https://github.com/microsoft/kiota-typescript/issues/1079

- Remove [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) evaluation to avoid situations where numbers like `0` or `false` are mistaken for null/undefined

### DIFF
--- a/packages/serialization/json/src/jsonSerializationWriter.ts
+++ b/packages/serialization/json/src/jsonSerializationWriter.ts
@@ -52,9 +52,10 @@ export class JsonSerializationWriter implements SerializationWriter {
       this.writer.push(JsonSerializationWriter.propertySeparator);
   };
   public writeNumberValue = (key?: string, value?: number): void => {
-    key && value && this.writePropertyName(key);
-    value && this.writer.push(`${value}`);
-    key && value && this.writer.push(JsonSerializationWriter.propertySeparator);
+    const isValuePresent = value !== null && value !== undefined;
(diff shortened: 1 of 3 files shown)


packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -52,9 +52,10 @@ export class JsonSerializationWriter implements SerializationWriter {
       this.writer.push(JsonSerializationWriter.propertySeparator);
   };
   public writeNumberValue = (key?: string, value?: number): void => {
-    key && value && this.writePropertyName(key);
-    value && this.writer.push(`${value}`);
-    key && value && this.writer.push(JsonSerializationWriter.propertySeparator);
+    const isValuePresent = value !== null && value !== undefined;
+    key && isValuePresent && this.writePropertyName(key);
+    isValuePresent && this.writer.push(`${value}`);
+    key && isValuePresent && this.writer.push(JsonSerializationWriter.propertySeparator);
```
andrueastman, line 58: Is it easier to handle it like this?
 ```suggestion
    if(isValuePresent){
      key && this.writePropertyName(key);
      this.writer.push(`${value}`);
      key && this.writer.push(JsonSerializationWriter.propertySeparator);
    }
```

andrueastman (approved):

andrueastman (approved):

Issue #1151, opened 2024-04-08
Cannot read properties of null (reading 'map')


illunix:
I use [v1.13.0](https://github.com/microsoft/kiota/releases/tag/v1.13.0)
Can someone help me resolve this, somehow Kiota can't handle null values
![image](https://github.com/microsoft/kiota/assets/42069493/656bd3e9-5304-499d-8f09-1cbd6268b232)
![image](https://github.com/microsoft/kiota/assets/42069493/52975354-3d87-4000-8ab1-d3640d2615c4)

baywet:
Hi @illunix 
Thanks for the additional information.
This is caused by a lack of defensive check here.
https://github.com/microsoft/kiota-typescript/blob/8c76a375bfab29d66e7ebe186f71342a1a2a4f54/packages/serialization/json/src/jsonParseNode.ts#L53

Would you be willing to submit a pull request to address this issue?

baywet:
more something like
```ts
if(!Array.isArray(this._jsonNode)) return undefined;
```

SrRickGrimes:
same issue here. 
I am getting this exception:

TypeError: Cannot read properties of null (reading 'map')
    at JsonParseNode.getCollectionOfPrimitiveValues

PR #1155, opened 2024-04-22
added validation check to see if the node is an Array

SrRickGrimes:
I was getting the same issue posted here.

fixes https://github.com/microsoft/kiota-typescript/issues/1151

I added a simple validation check to confirm the node is an Array.

### DIFF
--- a/packages/serialization/json/src/jsonParseNode.ts
+++ b/packages/serialization/json/src/jsonParseNode.ts
@@ -39,17 +39,21 @@ export class JsonParseNode implements ParseNode {
     typeof this._jsonNode === "object" &&
     (this._jsonNode as { [key: string]: any })[identifier] !== undefined
       ? new JsonParseNode(
-          (this._jsonNode as { [key: string]: any })[identifier],
+          (this._jsonNode as { [key: string]: any })[identifier]
         )
       : undefined;
   public getBooleanValue = () => this._jsonNode as boolean;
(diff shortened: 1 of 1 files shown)


SrRickGrimes:

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -82,15 +86,17 @@ export class JsonParseNode implements ParseNode {
     return undefined;
   }
   public getCollectionOfObjectValues = <T extends Parsable>(
-    method: ParsableFactory<T>,
+    method: ParsableFactory<T>
   ): T[] | undefined => {
-    return this._jsonNode ? (this._jsonNode as unknown[])
-      .map((x) => new JsonParseNode(x))
-      .map((x) => x.getObjectValue<T>(method)) : undefined;
+    return this._jsonNode
```
baywet, line 91: this spot also needs the check

baywet (changes requested):
Thanks for the contribution!
Added a comment.

packages/serialization/json/src/jsonParseNode.ts
SrRickGrimes: Thank you. I added a second validation. Please let me know if something else is missing.

baywet (approved):
thanks for making the changes!

PR #1204, opened 2024-05-30
Add a method to get the raw value of the node 

koros:
This will be used to support the generation of composed type which is made up of primitives:
``` Typescript
/**
 * Creates a new instance of the appropriate class based on discriminator value
 * @param parseNode The parse node to use to read the discriminator value and create the object
 * @returns {Primitives}
 */
export function createPrimitivesFromDiscriminatorValue(parseNode: ParseNode | undefined) : number | string | undefined {
    const nodeValue = parseNode?.getNodeValue();
    switch (typeof nodeValue) {
        case "number":
        case "string":
            return nodeValue;
        default:
            return undefined;
    }
}
```

### DIFF
--- a/packages/abstractions/src/serialization/parseNode.ts
+++ b/packages/abstractions/src/serialization/parseNode.ts
@@ -104,4 +104,9 @@ export interface ParseNode {
 	 * @return the byte array value of the node.
 	 */
 	getByteArrayValue(): ArrayBuffer | undefined;
+  /**
+	 * Gets the value for the current node
+	 * @return a parse node value for the current node element.
+	 */
+  getNodeValue(): unknown;
 }
--- a/packages/abstractions/src/serialization/parseNode.ts
+++ b/packages/abstractions/src/serialization/parseNode.ts
@@ -104,9 +104,4 @@ export interface ParseNode {
 	 * @return the byte array value of the node.
 	 */
 	getByteArrayValue(): ArrayBuffer | undefined;
-	/**
-	 * Gets the value for the current node
-	 * @return a parse node value for the current node element.
-	 */
-	getNodeValue(): unknown;
 }
--- a/packages/abstractions/src/serialization/parseNode.ts
+++ b/packages/abstractions/src/serialization/parseNode.ts
@@ -104,9 +104,9 @@ export interface ParseNode {
 	 * @return the byte array value of the node.
 	 */
 	getByteArrayValue(): ArrayBuffer | undefined;
-  /**
+	/**
 	 * Gets the value for the current node
 	 * @return a parse node value for the current node element.
 	 */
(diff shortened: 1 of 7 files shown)


packages/abstractions/src/serialization/parseNode.ts
```diff
@@ -104,4 +104,9 @@ export interface ParseNode {
 	 * @return the byte array value of the node.
 	 */
 	getByteArrayValue(): ArrayBuffer | undefined;
+	/**
+	 * Gets the value for the current node
+	 * @return a parse node value for the current node element.
+	 */
+	getNodeValue(): unknown;
```
andrueastman, line 111: I'm not sure if this is the right thing to do as the `parseNode` interface will no longer align with the implementation of the other languages.

andrueastman: Would it make useful to define two helper methods `IsNumberNode(ParseNode)` and `IsStringNode(ParseNode)` that essentially return true or false if the intent is to reduce generated code? This may also introduce type safety benefits rather than returning `unknown` types.

https://github.com/microsoft/kiota-typescript/blob/cd04c06ef4465a29164fc7d1eb9f3cd314d41609/packages/abstractions/src/serialization/untypedNumber.ts#L22

koros: Okay what would the complete code look like, for generating either a string | number for example

andrueastman: probably something like this? So that return values are also typed rather than returning `unknown`.

```suggestion
export function createPrimitivesFromDiscriminatorValue(parseNode: ParseNode | undefined) : number | string | undefined {
    if(IsNumberNode(parseNode)){
         return parseNode.getNumberValue();      
    }
     if(IsStringNode(parseNode)){
         return parseNode.getStringValue();      
    }
}
```

koros: The return value won't be unknows since the type definition has already been checked:
nodeValue

Your suggestion is giving an error confirm whether there's something I missed:
untypedNode

andrueastman: Just to confirm here, the suggestion would be to add methods like the ones I referenced. Not using them.

koros: I have refactored the generated code should look something like this:
``` Typescript
export function createPrimitivesFromDiscriminatorValue3(parseNode: ParseNode | undefined) : number | string | undefined {
  if (parseNode) {
    return parseNode.getNumberValue() || parseNode.getStringValue();
  }
  return undefined;
}
```

andrueastman (approved):

PR #1223, opened 2024-06-09
Feature/add tsc alias

koros:
Use tsc alias to add .js file extension to the ES bundled files to avoid import error when trying to use kiota in JavaScript environment


andrueastman (commented):
Any chance you can double check the failing CI?

koros:
Replaced by: https://github.com/microsoft/kiota-typescript/pull/1225

PR #1225, opened 2024-06-10
Feature/add tsc alias resolve js paths

koros:
Use tsc alias to add .js file extension to the ES bundled files to avoid import error when trying to use kiota in JavaScript environment


packages/authentication/azure/package.json
```diff
@@ -32,7 +32,7 @@
 	"homepage": "https://github.com/microsoft/kiota-typescript#readme",
 	"dependencies": {
 		"@azure/core-auth": "^1.5.0",
-		"@microsoft/kiota-abstractions": "*",
+		"@microsoft/kiota-abstractions": "file:../../abstractions/",
```
andrueastman, line 35: Out of curiosity, will the published package be affected by this? Should it point to a version range? Or this won't be breaking as it may look for file/folder?

packages/serialization/text/package.json
```diff
@@ -38,7 +38,7 @@
 	},
 	"homepage": "https://github.com/microsoft/kiota-typescript#readme",
 	"dependencies": {
-		"@microsoft/kiota-abstractions": "*",
+		"@microsoft/kiota-abstractions": "file:../../abstractions/",
```
andrueastman, line 41: Local ref is still present here

andrueastman (approved):

PR #1226, opened 2024-06-11
Handle Guid in query parameters.

SilasKenneth:
Fixes https://github.com/microsoft/kiota-typescript/issues/1113
closes #1116

### DIFF
--- a/packages/abstractions/test/common/requestInformation.ts
+++ b/packages/abstractions/test/common/requestInformation.ts
@@ -5,21 +5,12 @@
  * -------------------------------------------------------------------------------------------
  */
 
-import {assert, describe, it} from "vitest";
+import { assert, describe, it } from "vitest";
 
-import {
-  DateOnly,
-  HttpMethod,
-  type Parsable,
-  parseGuidString,
-  type RequestAdapter,
-  RequestInformation,
-  type SerializationWriter,
-  type SerializationWriterFactory, TimeOnly
-} from "../../src";
-import {MultipartBody} from "../../src/multipartBody";
-import {TestEnum} from "./store/testEnum";
-import {Guid} from "guid-typescript";
+import { DateOnly, HttpMethod, type Parsable, parseGuidString, type RequestAdapter, RequestInformation, type SerializationWriter, type SerializationWriterFactory, TimeOnly } from "../../src";
+import { MultipartBody } from "../../src/multipartBody";
+import { TestEnum } from "./store/testEnum";
+import { Guid } from "guid-typescript";
 
 interface GetQueryParameters {
 	select?: string[];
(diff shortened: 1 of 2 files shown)


packages/abstractions/test/common/requestInformation.ts
```diff
@@ -190,6 +192,15 @@ describe("RequestInformation", () => {
 		assert.equal(methodCalledCount, 1);
 	});
 
+	it("should correctly handle Guid type query parameter", () => {
+		const expected: string = `http://localhost/me?objectId=${parseGuidString("83afbf49-5583-152c-d7fb-176105d518bc")}`;
```
andrueastman, line 196:
```suggestion
		const expected: string = `http://localhost/me?objectId=83afbf49-5583-152c-d7fb-176105d518bc`;
```

packages/abstractions/test/common/requestInformation.ts
```diff
@@ -190,6 +192,15 @@ describe("RequestInformation", () => {
 		assert.equal(methodCalledCount, 1);
 	});
 
+	it("should correctly handle Guid type query parameter", () => {
```
andrueastman, line 195: Would it be possible to add tests for instances for `TimeOnly` and `DateOnly` types that could be in the URL as well? Similar to https://github.com/microsoft/kiota-abstractions-python/pull/278/files

SilasKenneth: This is fixed now @andrueastman

packages/abstractions/test/common/requestInformation.ts
```diff
@@ -190,6 +197,17 @@ describe("RequestInformation", () => {
...
 	});
 
+	it("should correctly handle Guid type query/path parameter", () => {
+		const expected: string = "http://localhost/users/33933a8d-32bb-c6a8-784a-f60b5a1dd66a/2021-12-12?objectId=83afbf49-5583-152c-d7fb-176105d518bc&startDate=2021-12-12&startTime=23%3A12%3A00.0000000&timeStamp=2024-06-10T21%3A00%3A00.000Z";
+		const requestInformation = new RequestInformation(HttpMethod.GET);
+		requestInformation.pathParameters["baseurl"] = baseUrl;
+		requestInformation.pathParameters["userId"] = Guid.parse("33933a8d-32bb-c6a8-784a-f60b5a1dd66a");
+		requestInformation.pathParameters["date"] = DateOnly.fromDate(new Date("2021-12-12"));
+		requestInformation.urlTemplate = "http://localhost/users/{userId}/{date}{?objectId,startDate,startTime,endDate,endTime,timeStamp}";
+		requestInformation.setQueryStringParametersFromRawObject<GetQueryParameters>({ objectId: parseGuidString("83afbf49-5583-152c-d7fb-176105d518bc"), startDate: new DateOnly({ year: 2021, month: 12, day: 12 }), startTime: new TimeOnly({ hours: 23, minutes: 12 }), timeStamp: new Date("2024-6-11") }, getQueryParameterMapper);
```
baywet, line 207: We have a timezone shift problem here, this is why the unit test is failing. You should set the timezone/offset as you create the date object.

baywet (changes requested):

packages/abstractions/test/common/requestInformation.ts
SilasKenneth: Thanks for the heads up @baywet.
I noticed that from the test logs and currently working on fixing it.

andrueastman (approved):

Issue #1188, opened 2024-05-15
Cannot write null to a nullable value


mbursill:
Our API depends on null's being allowed as part of the request:

```
const updateWidgetRequest: UpdateWidgetRequest = {
  name: 'Sproket',
  size: 5,
  inspectionDetails: null,
};
```

This is different than sending undefined for inspectionDetails. Our API treats undefined on patch as no change, but null as a means to clear the value.

While the swagger file defines the type as nullable, the generated code only allows `InspectionDetails | undefined`

We've tried to hack a value by doing:

`inspectionDetails: <InspectionDetails>(<unknown>null)`

This is awful and still doesn't work. This call to writeObjectValue strips it.

mbursill:
Tested with both backing store and without. The generated model is:

```
export interface UpdateWidgetRequest extends BackedModel, Parsable {
    /**
     * Stores model information.
     */
    backingStoreEnabled?: boolean;
    /**
     * The inspectionDetails property
     */
    inspectionDetails?: InspectionDetails;
    /**
     * The mass property
     */
    mass?: number;
    /**
     * The name property
     */
    name?: string;
    /**
     * The size property
     */
    size?: number;
}
```

The fields are all optional, meaning they're union with undefined, but assigning null is not possible.

mbursill:
> Is your goal of passing null to have null in the JSON payload to reset a field of an object for example?

Yes, that's exactly what we're wanting it for. Undefined doesn't work for us in this case. Our API uses undefined to mean a change to the field was not made, and not to alter the value. To clear/reset the value we need null.

rners01:
any updates on this?

baywet:
@rners01 you can run it straight in process if you want with a single command

```shell
dotnet run --project ./src/kiota/kiota.csproj -- arguments you usually pass to kiota
```

or you can tweak the .vscode/launch.json if you are using vs code.

Make sure you have [dotnet sdk 8 installed first](https://get.dot.net/)

Thanks for the time and help!

rners01:
but I don't see no `./src/kiota/kiota.csproj` in `kiota-typescript` repo
![image](https://github.com/microsoft/kiota-typescript/assets/115642795/0fb03926-7672-4018-8dd0-36bd989acb6b)

baywet:
sorry, I meant in [the main repo](https://github.com/microsoft/kiota)

rners01:
[...]

PR #1248, opened 2024-06-26
fix: process null values

Kindest13:
Fixes https://github.com/microsoft/kiota-typescript/issues/1188
1 part for processing null values


packages/abstractions/src/serialization/parseNode.ts
```diff
@@ -31,12 +31,12 @@ export interface ParseNode {
...
 	 * @return the boolean value of the node.
 	 */
-	getBooleanValue(): boolean | undefined;
+	getBooleanValue(): boolean | null | undefined;
 	/**
 	 * Gets the Number value of the node.
 	 * @return the Number value of the node.
 	 */
-	getNumberValue(): number | undefined;
+	getNumberValue(): number | null | undefined;
```
baywet, line 39: the other types also need to be changed

packages/abstractions/src/serialization/serializationWriter.ts
```diff
@@ -25,19 +25,19 @@ export interface SerializationWriter {
 	 * @param key the key to write the value with.
 	 * @param value the value to write to the stream.
 	 */
-	writeStringValue(key?: string | undefined, value?: string | undefined): void;
+	writeStringValue(key?: string | undefined, value?: string | null | undefined): void;
```
baywet, line 28: same comment about the other types

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -11,10 +11,10 @@ export class JsonParseNode implements ParseNode {
 	constructor(private readonly _jsonNode: unknown) {}
 	public onBeforeAssignFieldValues: ((value: Parsable) => void) | undefined;
 	public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
-	public getStringValue = () => (typeof this._jsonNode === "string" ? (this._jsonNode as string) : undefined);
+	public getStringValue = () => (typeof this._jsonNode === "string" || this._jsonNode == null  ? this._jsonNode : undefined);
```
baywet, line 14: shouldn't we use triple equals?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -11,10 +11,10 @@ export class JsonParseNode implements ParseNode {
...
 	public getChildNode = (identifier: string): ParseNode | undefined => (this._jsonNode && typeof this._jsonNode === "object" && (this._jsonNode as { [key: string]: any })[identifier] !== undefined ? new JsonParseNode((this._jsonNode as { [key: string]: any })[identifier]) : undefined);
-	public getBooleanValue = () => (typeof this._jsonNode === "boolean" ? (this._jsonNode as boolean) : undefined);
-	public getNumberValue = () => (typeof this._jsonNode === "number" ? (this._jsonNode as number) : undefined);
+	public getBooleanValue = () => (typeof this._jsonNode === "boolean" || this._jsonNode == null  ? this._jsonNode : undefined);
```
baywet, line 16: can you add unit tests for the null case please?

baywet (changes requested):
thanks for starting this. I have left a number of comments to move it forward.

packages/abstractions/src/serialization/parseNode.ts
Kindest13: changed

packages/abstractions/src/serialization/serializationWriter.ts
Kindest13: changed

packages/serialization/json/src/jsonParseNode.ts
Kindest13: agree, fixed

packages/serialization/json/src/jsonParseNode.ts
Kindest13: added

packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -75,6 +110,11 @@ export class JsonSerializationWriter implements SerializationWriter {
 		}
 	};
 	public writeCollectionOfObjectValues = <T extends Parsable>(key: string, values: T[], serializerMethod: ModelSerializerFunction<T>): void => {
```
baywet, line 112:
```suggestion
	public writeCollectionOfObjectValues = <T extends Parsable>(key: string, values: T[] | null | undefined, serializerMethod: ModelSerializerFunction<T>): void => {
```

(and repeat for other object/enum methods)

packages/serialization/json/src/jsonSerializationWriter.ts
```diff
@@ -214,7 +254,7 @@ export class JsonSerializationWriter implements SerializationWriter {
 		}
 		this.writer.push(JSON.stringify(value), JsonSerializationWriter.propertySeparator);
 	};
-	private readonly writeAnyValue = (key?: string | undefined, value?: unknown | undefined): void => {
+	private readonly writeAnyValue = (key?: string | undefined, value?: unknown | null | undefined): void => {
```
baywet, line 257: the conditions should be reviewed to write null only when the value is null, and not when it's undefined

Kindest13: should I write `undefined` when it's `undefined` or do nothing?

Kindest13: @baywet

packages/serialization/json/src/jsonSerializationWriter.ts
Kindest13: should I touch other places? `form`, `text`
image

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -11,15 +11,21 @@ export class JsonParseNode implements ParseNode {
...
-	public getStringValue = () => (typeof this._jsonNode === "string" ? (this._jsonNode as string) : undefined);
-	public getChildNode = (identifier: string): ParseNode | undefined => (this._jsonNode && typeof this._jsonNode === "object" && (this._jsonNode as { [key: string]: any })[identifier] !== undefined ? new JsonParseNode((this._jsonNode as { [key: string]: any })[identifier]) : undefined);
-	public getBooleanValue = () => (typeof this._jsonNode === "boolean" ? (this._jsonNode as boolean) : undefined);
-	public getNumberValue = () => (typeof this._jsonNode === "number" ? (this._jsonNode as number) : undefined);
-	public getGuidValue = () => parseGuidString(this.getStringValue());
-	public getDateValue = () => (this._jsonNode ? new Date(this._jsonNode as string) : undefined);
-	public getDateOnlyValue = () => DateOnly.parse(this.getStringValue());
-	public getTimeOnlyValue = () => TimeOnly.parse(this.getStringValue());
-	public getDurationValue = () => Duration.parse(this.getStringValue());
+	public getStringValue = () => (typeof this._jsonNode === "string" || this._jsonNode === null ? this._jsonNode : undefined);
```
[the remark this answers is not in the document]
Kindest13: basically I thought that this is the place where we do parsing of our json body so that's why added functionality to allow `null` values.
Is it the wrong assumption?

packages/serialization/json/src/jsonSerializationWriter.ts
baywet: Yes things need to be aligned between the interface season and the implementations.

packages/serialization/json/src/jsonParseNode.ts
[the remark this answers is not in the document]
baywet: Let's try to break this down. The properties of a model need to have null in the type definition so the user can set a null value. This also means that the argument of serialization writer need to accept null as a value so we can properly write it to the serialized representation.
Then we have mostly two choices:
- If the return type of parse node methods does not include null as a valid type it's going to be ok because it's still going to be a subset of what the property we're trying to assign accepts. During the deserialization we can treat null as functionally equivalent to undefined.
- If the return type of the parsnip methods does include null as a valid type then we need to make a difference between the serialization and return null.

The question then becomes do users care about telling the difference between null and undefined when it is returned by the service they are talking to? I would argue they do not since we've never had this request for any other language at this point. I would like to get you and @andrueastman opinion on the matter before we make a final decision on this.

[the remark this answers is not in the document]
andrueastman: > we've never had this request for any other language at this point.
Yeah. I would look at it through this lens. 
It would not be best placed to conceive of requirements here without a solid backing. The working default state as `undefined` would be okay for now and we could always change this once we get solid evidence of the need for distinguishing null and undefined for service level responses.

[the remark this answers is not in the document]
baywet: Thanks for chiming in here Andrew. 
@kindest13 are you clear on the modifications to be done at this point or do you have any additional questions?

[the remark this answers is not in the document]
Kindest13: As I understood there's no need to check on `null` value so just adding the `null` type is sufficient.
Although if `getStringValue` received `null` it returns `undefined` then how do we differenciate between `null` and `undefined`?

packages/serialization/json/test/common/testEntity.ts
```diff
@@ -11,17 +11,17 @@ import { Guid } from "guid-typescript";
...
 	additionalData?: Record<string, unknown>;
-	testDate?: Date | undefined;
+	testDate?: Date | null | undefined;
 	foos?: FooResponse[] | undefined;
-	id?: string | undefined;
-	testNumber?: number | undefined;
-	testGuid?: Guid | undefined;
+	id?: string | null | undefined;
+	testNumber?: number | null | undefined;
+	testGuid?: Guid | null | undefined;
```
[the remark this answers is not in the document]
Kindest13: which places are we talking about?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -53,7 +59,11 @@ export class JsonParseNode implements ParseNode {
 		}
 		return undefined;
 	}
-	public getCollectionOfObjectValues = <T extends Parsable>(method: ParsableFactory<T>): T[] | undefined => {
+	public getCollectionOfObjectValues = <T extends Parsable>(method: ParsableFactory<T>): T[] | null | undefined => {
+		if (this._jsonNode === null) {
+			return null;
+		}
+
```
Kindest13, line 66: what about these cases?
should I also remove `null` handling?
Will they handle it properly?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -11,15 +11,21 @@ export class JsonParseNode implements ParseNode {
...
+	public getDateValue = () => {
+		if (this._jsonNode === null) {
+			return null;
+		}
+
+		return this._jsonNode ? new Date(this._jsonNode as string) : undefined;
+	};
+	public getDateOnlyValue = () => (this.getStringValue() === null ? null : DateOnly.parse(this.getStringValue()));
+	public getTimeOnlyValue = () => (this.getStringValue() === null ? null : TimeOnly.parse(this.getStringValue()));
+	public getDurationValue = () => (this.getStringValue() === null ? null : Duration.parse(this.getStringValue()));
```
Kindest13, line 28: @baywet what about these cases?
should I also remove null handling?
Will they handle it properly?

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -53,7 +59,11 @@ export class JsonParseNode implements ParseNode {
 		}
 		return undefined;
 	}
-	public getCollectionOfObjectValues = <T extends Parsable>(method: ParsableFactory<T>): T[] | undefined => {
+	public getCollectionOfObjectValues = <T extends Parsable>(method: ParsableFactory<T>): T[] | null | undefined => {
+		if (this._jsonNode === null) {
+			return null;
```
Kindest13, line 64: @baywet  -//- https://github.com/microsoft/kiota-typescript/pull/1248/files#r1664428197

packages/serialization/json/src/jsonParseNode.ts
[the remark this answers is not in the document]
baywet: Let me reiterate so we make sure we're all on the same page. First, the models needs to be generated with the type, undefined, null. This is so people can assign null to properties. Then, the serialization writer method parameters need to be able to accept null as well. This is so we can write the new value to the serialized representation. Lastly, the parse node interface does not need to be updated but the implementation needs to treat no from a serialized value as it treats undefined today.
Does this clarify the changes that need to be made?

packages/serialization/json/src/jsonParseNode.ts
baywet: If we encounter null in a serialized response we need to treat it as undefined and return undefined instead.

packages/serialization/json/src/jsonParseNode.ts
baywet: same response

packages/serialization/json/test/common/testEntity.ts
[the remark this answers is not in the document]
Kindest13: @baywet  ?

packages/serialization/json/src/jsonSerializationWriter.ts
Kindest13: @baywet last on this, should I care about `multipart` serialization?
image

baywet: meaning: should you write a null value? or did you ask a different question?

Kindest13: yes and add `null` type

baywet: yes, I think we should both have null in the parameter type, and write null to json when the value is null (but not when it's undefined)

packages/serialization/json/test/common/testEntity.ts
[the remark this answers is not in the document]
baywet: form and multipart tests also have a test entity

packages/serialization/json/src/jsonSerializationWriter.ts
Kindest13: @baywet Should I touch the `writeByteArrayValue`?

image

Kindest13: @baywet I think I did it, please review
image

packages/abstractions/src/serialization/kiotaJsonSerializer.ts
```diff
@@ -56,7 +56,7 @@ export function serializeCollectionToJsonAsString<T extends Parsable>(values: T[
  * @param factory the factory for the model type
  * @returns the deserialized parsable object
  */
-export function deserializeFromJson<T extends Parsable>(bufferOrString: ArrayBuffer | string, factory: ParsableFactory<T>): Parsable {
+export function deserializeFromJson<T extends Parsable>(bufferOrString: ArrayBuffer | string, factory: ParsableFactory<T>): Parsable | null {
```
baywet, line 59: I believe this file needs to be reverted

packages/abstractions/src/serialization/kiotaSerializer.ts
```diff
@@ -85,7 +85,7 @@ function getStringValueFromBuffer(buffer: ArrayBuffer): string {
  * @param factory the factory for the model type
  * @returns the deserialized parsable object
  */
-export function deserialize<T extends Parsable>(contentType: string, bufferOrString: ArrayBuffer | string, factory: ParsableFactory<T>): Parsable {
+export function deserialize<T extends Parsable>(contentType: string, bufferOrString: ArrayBuffer | string, factory: ParsableFactory<T>): Parsable | null {
```
baywet, line 88: I believe this file needs to be reverted

packages/abstractions/src/serialization/parseNode.ts
```diff
@@ -20,75 +20,75 @@ export interface ParseNode {
 	 * Gets the string value of the node.
 	 * @return the string value of the node.
 	 */
-	getStringValue(): string | undefined;
+	getStringValue(): string | null | undefined;
```
baywet, line 23: given our latest discussions, I believe this file needs to be reverted.

packages/serialization/form/src/formParseNode.ts
```diff
@@ -36,7 +36,7 @@ export class FormParseNode implements ParseNode {
 	public onBeforeAssignFieldValues: ((value: Parsable) => void) | undefined;
 	public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
 	public getStringValue = (): string => decodeURIComponent(this._rawString);
-	public getChildNode = (identifier: string): ParseNode | undefined => {
+	public getChildNode = (identifier: string): ParseNode | null | undefined => {
```
baywet, line 39: this file needs to be reverted as well

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -12,15 +12,15 @@ export class JsonParseNode implements ParseNode {
 	public onBeforeAssignFieldValues: ((value: Parsable) => void) | undefined;
 	public onAfterAssignFieldValues: ((value: Parsable) => void) | undefined;
```
baywet, line 13: the prototype changes on this file need to be reverted

packages/serialization/json/test/common/JsonParseNode.ts
```diff
@@ -70,16 +70,16 @@ describe("JsonParseNode", () => {
 		assert.equal(enumValueResult, TestEnumObject.A);
 	});
 
-	it("Test a null collection of object values", async () => {
+	it("Test an undefined collection of object values", async () => {
```
baywet, line 73: we also need a test for null

packages/serialization/text/src/textParseNode.ts
```diff
@@ -17,7 +17,7 @@ export class TextParseNode implements ParseNode {
 			this.text = this.text.substring(1, this.text.length - 2);
 		}
 	}
-	public getByteArrayValue(): ArrayBuffer | undefined {
+	public getByteArrayValue(): ArrayBuffer | null | undefined {
```
baywet, line 20: this file needs to be reverted

packages/serialization/text/src/textSerializationWriter.ts
```diff
@@ -19,53 +19,77 @@ export class TextSerializationWriter implements SerializationWriter {
 		this.writeStringValue(key, b64);
 	}
 	private static noStructuredDataMessage = "text does not support structured data";
-	private readonly writer: string[] = [];
+	private readonly writer: (string | null)[] = [];
```
baywet, line 22: why not simply write the null text representation and keep it as an array of strings?

baywet (changes requested):
I think we're making good progress here! Thanks for the continuous efforts.

packages/serialization/json/test/common/JsonParseNode.ts
Kindest13: @baywet `null`, also returns `undefined` , right?

Kindest13: image

baywet: > @baywet `null`, also returns `undefined` , right?

correct.

I'm not sure I understand the second comment though?

baywet (commented):
I've just pushed a set of changes to fix the unit test for form serialization writer. I think similar fixes are required for the json serialization writer as its test is also failing.

Kindest13:
> I've just pushed a set of changes to fix the unit test for form serialization writer. I think similar fixes are required for the json serialization writer as its test is also failing.

@baywet added

packages/serialization/json/test/common/JsonParseNode.ts
Kindest13: nvm, it's just the screenshot of `null` case test

baywet (dismissed):
LGTM!
@andrueastman for a final review

Kindest13:
> LGTM! @andrueastman for a final review

Great! I just comitted changes to remove `.only` from test file

baywet (dismissed):

andrueastman (approved):
LGTM.

@Kindest13 I believe you still need to reply to the policy service bot to get this merged.

rners01:
> @Kindest13 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

Kindest13:
> @Kindest13 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

baywet:
Now this PR is ready, I'd like us to get started on the generation updates before we merge this pull request.
@Kindest13 do you have all the information you need to get started in the kiota repository or do you want additional pointers?

Kindest13:
> Now this PR is ready, I'd like us to get started on the generation updates before we merge this pull request. @Kindest13 do you have all the information you need to get started in the kiota repository or do you want additional pointers?

Would be great if you can share more info on this

baywet:
My guess is that since the `?` is hardcoded here. And since only the properties need to be settable to null, we could probably hardcoded it here.
https://github.com/microsoft/kiota/blob/4d59461de79d1de58a309b4c25fee7b005a15ee5/src/Kiota.Builder/Writers/TypeScript/CodePropertyWriter.cs#L37

Let us know if you have additional questions.

baywet:
` | null` roughly

rners01:
> hardcoded

Hey @baywet, is that correct?
So just this line?
`writer.WriteLine($"{codeElement.Name.ToFirstCharacterLowerCase()}?: {returnType}{(isFlagEnum ? "[]" : string.Empty)} | null;");`

Kindest13:
> ` | null` roughly

Hey @baywet, is that correct?
So just this line?
`writer.WriteLine($"{codeElement.Name.ToFirstCharacterLowerCase()}?: {returnType}{(isFlagEnum ? "[]" : string.Empty)} | null;");`

baywet:
yes it looks correct, however I left a couple of comments on the PR to wrap it up.

baywet (approved):

Issue #1260, opened 2024-07-05
[TS] Rollout a Kiota.Bundle package


andrueastman:
- [ ] Add a new package/project for the Kiota.Bundle that pulls/depends in the abstractions libraries. 
- [ ] Expose a derived request adapter to auto-register the serialization pieces.

Related to https://github.com/microsoft/kiota/issues/4636

PR #1298, opened 2024-07-30
Setup bundle for TS

andrueastman:
Closes https://github.com/microsoft/kiota-typescript/issues/1260


baywet (dismissed):

baywet:
@andrueastman with the way the publishing works, you might want to bump versions for other packages so we don't get a failure of the version already being published.

Issue #1299, opened 2024-07-30
Proxifier does not validate for missing path parameters 


baywet:
Source of the issue is here. 
https://github.com/microsoft/kiota-typescript/blob/e8317bbcf714c487cccaa2ba5d965ec71b3f019c/packages%2Fabstractions%2Fsrc%2FapiClientProxifier.ts#L197

For a full repro, check the linked issue below

PR #1322, opened 2024-08-13
Validate missing path params 

andrueastman:
Closes https://github.com/microsoft/kiota-typescript/issues/1299


packages/abstractions/src/apiClientProxifier.ts
```diff
@@ -194,7 +194,11 @@ export const apiClientProxifier = <T extends object>(requestAdapter: RequestAdap
 						if (navigationCandidate.pathParametersMappings && navigationCandidate.pathParametersMappings.length > 0) {
 							for (let i = 0; i < argArray.length; i++) {
 								const element = argArray[i];
-								downWardPathParameters[navigationCandidate.pathParametersMappings[i]] = element;
+								if (element) {
```
baywet, line 197:
```suggestion
								if (element === undefined || element === null || element === '') {
```
In case the parameter is a number and somehow there's something special with id zero or something

baywet (changes requested):
Thanks for taking this one on!

baywet (approved):

PR #1358, opened 2024-09-10
feat: request body compression

rkodev:
Resolves https://github.com/microsoft/kiota-typescript/issues/1288

Adds support for compression of request body


packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+
+		// execute the next middleware and check if the response code is 415
+		try {
+			const response = await this.next?.execute(url, requestInit, requestOptions);
+			if (response?.status === 415) {
+				// remove the Content-Encoding header
+				requestInit.headers.delete("Content-Encoding");
+				requestInit.body = unCompressedBody.buffer;
+				span?.setAttribute("http.request_body_compressed", false);
+				span?.setAttribute("http.request_content_length", unCompressedBodySize);
```
baywet, line 84: same comments about updating the attribute names

packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+			return this.next?.execute(url, requestInit, requestOptions) ?? Promise.reject(new Error("Response is undefined"));
+		}
+	}
+
+	private contentRangeBytesIsPresent(header: HeadersInit | undefined): boolean {
+		if (!header) {
+			return false;
+		}
+		if (header instanceof Headers) {
+			const contentRange = header.get("Content-Range");
```
baywet, line 98: define a constant for content-range instead of duplicating it

baywet (changes requested):
Great work getting started on this!

rkodev:
@baywet I think we have a bit off on issue with compression. Native support for compression in node for the web is a bit on an issue, as of node 22 [zlib ](https://nodejs.org/api/zlib.html) has native support and it works great, but for node 18 / 20 we do need to import it to support browser compression.

baywet:
@rkodev the compression stream API is available on 90+% of browsers out there, we should use that for browsers and assume if applications run into an issue they can selectively disable the middleware based on the browser they want to support.
https://caniuse.com/?search=compressionstream
https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API

As for nod ethe zlib gzip API is available since "forever"
https://nodejs.org/api/zlib.html#zlibgzipbuffer-options-callback https://nodejs.org/api/zlib.html#zlibcreategzipoptions

The way around the issue you're facing is to:
1. have two implementation methods of the compression (one for browser, one for node)
2. leverage dynamic imports in those methods (and not at the top of the document)
3. selectively use the right method based on the context (node or browser)

I think we have other examples of doing that across the code.

If this is two complex, we could have two implementations of the middleware (a bit more duplication) and insert the right implementation in the right factory (we already have that in place as well)

Let me know if you have any additional comments or questions.

packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,187 @@
...
+		}
+		if (header instanceof Headers) {
+			return header.has(CompressionHandler.CONTENT_ENCODING_HEADER);
+		} else if (typeof header === "object") {
+			return CompressionHandler.CONTENT_ENCODING_HEADER in header;
+		}
+		return false;
+	}
+
+	private async readBodyAsBytes(body: BodyInit | null | undefined): Promise<Uint8Array> {
```
calebkiage, line 135: Shouldn't the types you're testing against be included in the union?

packages/http/fetch/src/middlewares/options/compressionHandlerOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+	 * @inheritdoc
+	 */
+	public getKey(): string {
+		return CompressionHandlerOptionsKey;
+	}
+	/**
+	 * Returns whether the compression handler is enabled or not.
+	 * @returns whether the compression handler is enabled or not.
+	 */
+	public get ShouldCompress(): boolean {
```
calebkiage, line 41:
```suggestion
	public get shouldCompress(): boolean {
```

Isn't the convention in typescript camelCase?

baywet (dismissed):

packages/http/fetch/src/middlewares/browser/middlewareFactory.ts
```diff
@@ -25,6 +26,6 @@ export class MiddlewareFactory {
 	 */
 	public static getDefaultMiddlewares(customFetch: (request: string, init: RequestInit) => Promise<Response> = (...args) => fetch(...args) as any): Middleware[] {
 		// Browsers handles redirection automatically and do not require the redirectionHandler
-		return [new RetryHandler(), new ParametersNameDecodingHandler(), new UserAgentHandler(), new HeadersInspectionHandler(), new CustomFetchHandler(customFetch)];
+		return [new RetryHandler(), new ParametersNameDecodingHandler(), new UserAgentHandler(), new CompressionHandler(), new HeadersInspectionHandler(), new CustomFetchHandler(customFetch)];
```
baywet, line 29: now that the middleware is compatible with both node and browser, we should also add it to the node factory

baywet (approved):

baywet (approved):

rkodev:
@baywet Thanks for taking a look

packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+				compressedBody,
+				size: compressedBody.length,
+			};
+		}
+	}
+
+	private async compressUsingZlib(body: unknown): Promise<Buffer> {
+		// @ts-ignore
+		const zlib = await import("zlib");
+		return new Promise((resolve, reject) => {
```
baywet, line 199: We might need an additional await here for consistency, please double check

packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+					reject(err);
+				} else {
+					resolve(compressed);
+				}
+			});
+		});
+	}
+
+	private async compressUsingCompressionStream(uint8ArrayStream: ReadableStream<Uint8Array>): Promise<{ body: ArrayBuffer; size: number }> {
+		const compressionStream = new CompressionStream("gzip");
```
baywet, line 211: Should we rely on a dynamic import for compression stream as well?

packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -0,0 +1,239 @@
...
+					return this.executeInternal(currentOptions, url, requestInit as FetchRequestInit, requestOptions, span);
+				} finally {
+					span.end();
+				}
+			});
+		}
+		return this.executeInternal(currentOptions, url, requestInit as FetchRequestInit, requestOptions);
+	}
+
+	private async executeInternal(options: CompressionHandlerOptions, url: string, requestInit: FetchRequestInit, requestOptions?: Record<string, RequestOption> | undefined, span?: Span): Promise<Response> {
```
andrueastman, line 72:
```suggestion
	private async executeInternal(options: CompressionHandlerOptions, url: string, requestInit: FetchRequestInit, requestOptions?: Record<string, RequestOption>, span?: Span): Promise<Response> {
```
parameter is already optional..

andrueastman (commented):
Please also add a config to remove this line as well. (I should have used a different approach to set a marker for the release) 

https://github.com/microsoft/kiota-typescript/blob/255b504a0ec9036d47eac8188bc72252bf762b88/release-please-config.json#L11

packages/http/fetch/src/middlewares/compressionHandler.ts
rkodev: @baywet the  `CompressionStream`  is part of TS `lib.dom`, there isn't any import statement for compression stream

packages/http/fetch/src/middlewares/compressionHandler.ts
rkodev: @baywet could you clarify this abit

baywet: we're returning the promise without awaiting it line 199. I think that TypeScript/JavaScript kind of "gets" what we're trying to do since promise chaining back in the days also accepted a returned promise. But it might be worth awaiting it as well to ensure things are properly chained since we're awaiting another promise above.

baywet (approved):

jlarmstrongiv:
It took me a while to track this bug down!  Enabling request body compression by default broke several of my clients.  Could it be an optional middleware that’s not enabled by default?

baywet:
@jlarmstrongiv thanks for reporting this. Can you share more information as to how were your clients brokens? Did the target service not support decompress request bodies? or was something wrong with how the middleware handles the compression/headers?

The design is to enable it by default across languages, of course you always have the option to disable it for a specific client.

Additionally, was your client running as a node or a browser application?

@rkodev to follow up on that as I'll be out of office soon.

jlarmstrongiv:
@baywet @rkodev

While the kiota client supports compressed request bodies just fine, the API I use does not.  Instead, the [`@microsoft/kiota-http-fetchlibrary`](https://www.npmjs.com/package/@microsoft/kiota-http-fetchlibrary) transitive dependency updated without me knowing it, and broke my app.  The server in question responds with 500 errors when 'Content-Encoding': 'gzip' bodies are found in the request.

I would like to avoid having the request compression middleware enabled by default on all languages and avoid running into this issue 5 more times.  Or, at least, have an example on how to disable the middleware for each language.

I run the client as both a Node.js and Browser application.  It’s the server that has the issue, and it’s a server I do not have control over.  Without the request content encoding, everything works great.

baywet:
Thank you for the additional information.

Is this server not supporting compression for all operations/paths or just for a few?

The reason why I'm asking is because you can pass a [per request option](https://github.com/microsoft/kiota-typescript/blob/e8b88a36faa320d7cb40984fd645a7efecb07586/packages/http/fetch/src/middlewares/options/compressionHandlerOptions.ts#L22) if this is just for a few to disable body compression on a per request base.

Otherwise your best bet is to copy the middleware and client factory to tweak which handlers are selected before passing things to the request adapter.

This advice applies to all languages as this is the general design of the middleware + options + client part of the infrastructure.

As for the fact that the compression handler has been added in a preview bump, while this is an important behavior change, it's not an API breaking change, and TypeScript is still in preview anyway, be aware that other changes of the sort might happen before the language reaches a stable maturity.

jlarmstrongiv:
Unfortunately, it’s all operations.  I tried the following code:

```ts
const httpClient = new HttpClient(
    undefined,
    new RetryHandler(),
    new RedirectHandler(),
    new ParametersNameDecodingHandler(),
    new UserAgentHandler(),
    new HeadersInspectionHandler(),
  );
const fetchAdapter = new FetchRequestAdapter(
    authenticationProvider,
    undefined,
    undefined,
    httpClient,
  );
```

But I receive the error `next middleware is undefined.  `.  Do you have an example for disabling compression for all operations/paths?  It’s not immediately clear how to do so.

> As for the fact that the compression handler has been added in a preview bump, while this is an important behavior change, it's not an API breaking change, and TypeScript is still in preview anyway, be aware that other changes of the sort might happen before the language reaches a stable maturity.

I have subscribed to all issues related to request compression to ensure I am notified when breaking changes occur.  In the several months I’ve worked with kiota, this issue is the first to have caused outages.

baywet:
Thank you for the additional information.

Can you try the following? It's not the most obvious way of setting it up.

```diff
const httpClient = new HttpClient(
-   undefined,
+   fetch, 
    new RetryHandler(),
    new RedirectHandler(),
    new ParametersNameDecodingHandler(),
    new UserAgentHandler(),
    new HeadersInspectionHandler(),
  );
```

[see the implementation](https://github.com/microsoft/kiota-typescript/blob/e8b88a36faa320d7cb40984fd645a7efecb07586/packages/http/fetch/src/httpClient.ts#L35)

PR #1359, opened 2024-09-10
Update retry attribute names

rkodev:
Resolves https://github.com/microsoft/kiota-typescript/issues/1318

Update retry attribute and implement resend delay


packages/http/fetch/src/middlewares/retryHandler.ts
```diff
@@ -165,14 +165,18 @@ export class RetryHandler implements Middleware {
...
 			}
 			if (tracerName) {
 				return await trace.getTracer(tracerName).startActiveSpan(`retryHandler - attempt ${retryAttempts}`, (span) => {
 					try {
-						span.setAttribute("http.retry_count", retryAttempts);
+            span.setAttribute("http.request.resend_count", retryAttempts);
+            if(delay){
+              span.setAttribute("http.request.resend_delay", delay);
+            }
 						span.setAttribute("http.status_code", response.status);
```
baywet, line 180:
```suggestion
						span.setAttribute("http.response.status_code", response.status);
```

(there are probably other places)

PR #1423, opened 2024-10-16
Fix malformed multipart body

shemogumbe:
Opened to replace https://github.com/microsoft/kiota-typescript/pull/1069 which seems to be from a fork I can't access

Fixes https://github.com/microsoft/kiota-typescript/issues/1068


packages/abstractions/src/multipartBody.ts
```diff
@@ -93,7 +93,7 @@ interface MultipartEntry {
 	serializationCallback?: ModelSerializerFunction<Parsable>;
 }
 
-export const serializeMultipartBody = (writer: SerializationWriter, multipartBody: Partial<MultipartBody> | undefined | null = new MultipartBody()): void => {
```
baywet, line 96: please revert the const changes on this line. Here is more context https://www.freecodecamp.org/news/the-difference-between-arrow-functions-and-normal-functions/

packages/serialization/multipart/src/multipartSerializationWriter.ts
```diff
@@ -30,51 +30,48 @@ export class MultipartSerializationWriter implements SerializationWriter {
 	public onBeforeObjectSerialization: ((value: Parsable) => void) | undefined;
 	public onAfterObjectSerialization: ((value: Parsable) => void) | undefined;
 	public onStartObjectSerialization: ((value: Parsable, writer: SerializationWriter) => void) | undefined;
-	public writeStringValue = (key?: string, value?: string | null): void => {
+	public writeStringValue = (key?: string, value?: string): void => {
```
baywet, line 33: please revert the removal of null here as it does not conform with the interface

packages/serialization/multipart/src/multipartSerializationWriter.ts
```diff
@@ -30,51 +30,48 @@ export class MultipartSerializationWriter implements SerializationWriter {
...
-
-			this.writeByteArrayValue(undefined, new TextEncoder().encode(isNullValue ? "null" : value).buffer);
+	private writeRawStringValue = (value?: string): void => {
+		if (value) {
+			this.writeByteArrayValue(undefined, new TextEncoder().encode(value).buffer);
 		}
 	};
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	public writeBooleanValue = (key?: string, value?: boolean | null): void => {
+	public writeBooleanValue = (key?: string, value?: boolean): void => {
```
baywet, line 50: same comment about the null option for all the following methods

packages/abstractions/src/multipartBody.ts
```diff
@@ -114,16 +114,20 @@ export const serializeMultipartBody = (writer: SerializationWriter, multipartBod
 	let first = true;
 	for (const partName in parts) {
 		if (Object.prototype.hasOwnProperty.call(parts, partName)) {
+			// Added guard
 			if (first) {
 				first = false;
 			} else {
-				writer.writeStringValue(undefined, "");
+				writer.writeStringValue(undefined, "\r\n");
```
baywet, line 121: can we add a unit test to prevent future regressions?

baywet (changes requested):

packages/abstractions/src/multipartBody.ts
shemogumbe: removed

Isn't this what we moved from?

packages/serialization/multipart/src/multipartSerializationWriter.ts
```diff
@@ -34,47 +34,44 @@ export class MultipartSerializationWriter implements SerializationWriter {
...
+			this.writeByteArrayValue(undefined, new TextEncoder().encode(value).buffer);
 		}
 	};
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	public writeBooleanValue = (key?: string, value?: boolean | null): void => {
 		throw new Error(`serialization of boolean values is not supported with multipart`);
 	};
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	public writeNumberValue = (key?: string, value?: number | null): void => {
+	public writeNumberValue = (key?: string, value?: number): void => {
```
andrueastman, line 54: I believe all the `writeXXXXValue` methods should be reverted to include the `| null` in the parameter type. Otherwise they will no longer conform to the definitions defined in the `SerializationWriter` interface. 
https://github.com/microsoft/kiota-typescript/blob/434513968aa865908809a488f8b84200486a4d6f/packages/abstractions/src/serialization/serializationWriter.ts#L40

andrueastman (commented):
Looks good from my end. 
I believe you'll need to add at least one conventional commit to trigger release and update the changelog with release please.

andrueastman (approved):

packages/abstractions/src/multipartBody.ts
```diff
@@ -93,7 +93,7 @@ interface MultipartEntry {
 	serializationCallback?: ModelSerializerFunction<Parsable>;
 }
 
-export const serializeMultipartBody = (writer: SerializationWriter, multipartBody: Partial<MultipartBody> | undefined | null = new MultipartBody()): void => {
+export function serializeMultipartBody(writer: SerializationWriter, multipartBody: Partial<MultipartBody> = new MultipartBody()): void {
```
baywet, line 96:
```suggestion
export const serializeMultipartBody = (writer: SerializationWriter, multipartBody: Partial<MultipartBody> = new MultipartBody()): void => {
```

baywet (changes requested):

baywet (dismissed):

@shemogumbe please make sure you squash merge (not merge) with a conventional commit to trigger the release.

baywet (approved):

PR #1399, opened 2024-10-02
Fixes `serializeMultipartBody` function signature.

andrueastman:
Related to https://github.com/microsoft/kiota/pull/5518

The `serializeMultipartBody` function does not match the `ModelSerializerFunction<Parsable>` type after updates made in https://github.com/microsoft/kiota-typescript/pull/1248

PR #1435, opened 2024-10-22
fix: add multipart to request body serializers

rkodev:
minor fix: adds Multipart type to options for setContentFromParsable.

### DIFF
--- a/packages/abstractions/src/apiClientProxifier.ts
+++ b/packages/abstractions/src/apiClientProxifier.ts
@@ -10,6 +10,7 @@ import type { ErrorMappings, PrimitiveTypesForDeserialization, PrimitiveTypesFor
 import type { RequestConfiguration } from "./requestConfiguration";
 import { RequestInformation, type RequestInformationSetContent } from "./requestInformation";
 import type { ModelSerializerFunction, Parsable, ParsableFactory } from "./serialization";
+import { MultipartBody } from "./multipartBody";
 const sanitizeMethodName = (methodName: string): string => {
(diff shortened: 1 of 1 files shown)


packages/abstractions/src/apiClientProxifier.ts
```diff
@@ -216,7 +217,7 @@ export interface RequestMetadata {
 	errorMappings?: ErrorMappings;
 	adapterMethodName?: SendMethods;
 	responseBodyFactory?: ParsableFactory<Parsable> | PrimitiveTypesForDeserialization;
-	requestBodySerializer?: ModelSerializerFunction<Parsable> | PrimitiveTypesForDeserialization;
+	requestBodySerializer?: ModelSerializerFunction<Parsable> | ModelSerializerFunction<MultipartBody> | PrimitiveTypesForDeserialization;
```
baywet, line 220: since [multipartBody implements IParsable](https://github.com/microsoft/kiota-typescript/blob/1edca86f2e3e9007cfe6f132e28503850d42d4f3/packages/abstractions/src/multipartBody.ts#L14) what does this solve?

andrueastman: +1 on the comment by @baywet 

There was an inconsistency with the type information earlier releases but that was resolved with the PR at https://github.com/microsoft/kiota-typescript/pull/1399 as the function signature was missing the null information.

rkodev: The build error was as a result of an older version

PR #1475, opened 2024-11-05
fix: getting rid of unnecessary guid-typescript dependency

markovav-official:
Fixing PR [microsoft/kiota-typescript#1154](https://github.com/microsoft/kiota-typescript/pull/1154)

Fixes [microsoft/kiota#4508](https://github.com/microsoft/kiota/issues/4508)
Using a custom Guid type to get rid of unnecessary package (and partly broken) guid-typescript.

PR #1481, opened 2024-11-11
fix: removed export for guid part generation

markovav-official:
Accepted suggestion from [microsoft/kiota-typescript#1475](https://github.com/microsoft/kiota-typescript/pull/1475)


baywet (approved):

PR #1489, opened 2024-11-18
fix: update std-uritemplate to v2.0.0

### LINKED ISSUE #1384: prepare for std uri template 2 and date time drop
related https://github.com/std-uritemplate/std-uritemplate/pull/245
The following types need to the normalized to their RFC3339 representation before they are passed to std URI template when they are present are query or path parameter values
Date their collection representations (array, List)

musale:
Closes #1384 

Updates the std-uritemplate package to v2.0.0. No further code changes required
https://github.com/microsoft/kiota-typescript/blob/7643e7c49b9318e74359f462e8063a4c7861efe0/packages/abstractions/src/requestInformation.ts#L242-L253

and passing tests

https://github.com/microsoft/kiota-typescript/blob/7643e7c49b9318e74359f462e8063a4c7861efe0/packages/abstractions/test/common/requestInformation.ts#L223-L232

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3528,9 +3528,10 @@
       "dev": true
     },
     "node_modules/@std-uritemplate/std-uritemplate": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/@std-uritemplate/std-uritemplate/-/std-uritemplate-1.0.6.tgz",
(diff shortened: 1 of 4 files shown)


baywet (changes requested):
Thanks for the contribution!

From the quoted normalization, I think we're missing duration. Would you mind adding it please?

baywet (approved):

PR #1495, opened 2024-11-20
feat(abstractions): create empty guid 

illunix:
After update to 1.20.0 I saw that kiota don't use guid-typescript anymore and my code base is now missing method for generating empty guid


baywet (approved):
Thanks for the contribution!

Issue #1442, opened 2024-10-26
kiota with angular error


marcioatws:
Sorry for my english, it's not my native language.

kiota version: `1.19.1+d294e04ba7f756896878a7015df1648f9dc0bcde`
kiota bundle version: `1.0.0-preview.70`

angular cli: `18.2.10`
angular dev kit: `18.2.10`

angular builder configuration:
production: `"builder": "@angular-devkit/build-angular:application"`
development: `"builder": "@angular-devkit/build-angular:dev-server"`

tsconfig

```
/* To learn more about this file see: https://angular.io/config/tsconfig. */
{
  "angularCompilerOptions": {
    "enableI18nLegacyMessageIdFormat": false,
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true,
    "strictTemplates": true
  },
  "compileOnSave": false,
  "compilerOptions": {
    "declaration": false,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true,
    "importHelpers": true,
    "isolatedModules": true,
    "lib": [
      "ESNext",
      "dom"
    ],
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "noFallthroughCasesInSwitch": true,
    "noImplicitOverride": true,
    "noImplicitReturns": true,
    "noPropertyAccessFromIndexSignature": true,
    "outDir": "./dist/out-tsc",
    "paths": {

    },
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "target": "ES2022",
    "useDefineForClassFields": false,
  },
}
```

I can run in dev without errors, but when i try run production:

```
X [ERROR] Could not resolve "zlib"

    node_modules/@microsoft/kiota-http-fetchlibrary/dist/es/src/middlewares/compressionHandler.js:176:32:
      176 │       const zlib = await import("zlib");
          ╵                                 ~~~~~~

  The package "zlib" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
```

and

```
X [ERROR] TS2322: Type '(writer: SerializationWriter, multipartBody?: Partial<MultipartBody> | undefined) => void' is not assignable to type 'PrimitiveTypesForDeserialization | ModelSerializerFunction<Parsable> | undefined'.
[...]

PR #1499, opened 2024-11-22
fix: remove zlib compression in node

rkodev:
Resolve https://github.com/microsoft/kiota-typescript/issues/1442

Removes support for zlib in compression handler (in node)

### DIFF
--- a/packages/http/fetch/src/middlewares/compressionHandler.ts
+++ b/packages/http/fetch/src/middlewares/compressionHandler.ts
@@ -169,37 +169,13 @@ export class CompressionHandler implements Middleware {
 		compressedBody: ArrayBuffer | Buffer;
 		size: number;
 	}> {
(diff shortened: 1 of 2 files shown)


packages/http/fetch/src/middlewares/compressionHandler.ts
```diff
@@ -169,37 +169,13 @@ export class CompressionHandler implements Middleware {
...
-		return await new Promise((resolve, reject) => {
-			zlib.gzip(body as string | ArrayBuffer | NodeJS.ArrayBufferView, (err, compressed) => {
-				if (err) {
-					reject(err);
-				} else {
-					resolve(compressed);
-				}
-			});
-		});
+		// in browser
```
baywet, line 172:
```suggestion
```

baywet (changes requested):

Issue #1518, opened 2024-12-04
http - add authorization middleware support and factory overload


baywet:
More context https://github.com/microsoft/kiota-dotnet/issues/270
Specification https://github.com/microsoftgraph/msgraph-sdk-design/blob/main/middleware/AuthorizationHandler.md

Once implemented, add a new issue in microsoft/msgraph-sdk-typescript-core to add the overloads there as well

PR #1575, opened 2025-01-28
feat: adds authorization handler

rkodev:
Resolves https://github.com/microsoft/kiota-typescript/issues/1518

Add authorization handler middleware


packages/http/fetch/src/middlewares/authorizationHandler.ts
```diff
@@ -0,0 +1,139 @@
...
+				requestInformation.headers.add(key, value);
+			}
+		}
+
+		// add all the headers to fetchRequestInit
+		if (!fetchRequestInit.headers) {
+			fetchRequestInit.headers = {};
+		}
+
+		await this.authenticateRequest(requestInformation);
```
baywet, line 68: we should instead add a getter for the access token provider on the authentication provider and just pass the URI

baywet (changes requested):
thanks for getting this started!

packages/http/fetch/src/middlewares/authorizationHandler.ts
rkodev: had to change the constructor to use `AccessTokenProvider` directly

baywet: sorry if I wasn't clear, the constructor should still accept a base bearer auth provider because this is what the user will have at hand. But that class should expose a property to let us read the access token provider so we can use it here.
Does this make things a bit more clear?

rkodev: makes sense

baywet: while I'm at it, we'll need an optional parameter for a base bearer auth provider in the kiota client factory, and for this middleware to be part of the default factory.

rkodev: Made both changes

packages/abstractions/src/authentication/baseBearerTokenAuthenticationProvider.ts
```diff
@@ -35,4 +35,6 @@ export class BaseBearerTokenAuthenticationProvider implements AuthenticationProv
 			}
 		}
 	};
+
+	public getAccessTokenProvider = () => this.accessTokenProvider;
```
baywet, line 39: can't we use a getter instead?

baywet (commented):
One remaining question on my side.

packages/abstractions/src/authentication/baseBearerTokenAuthenticationProvider.ts
```diff
@@ -35,4 +35,8 @@ export class BaseBearerTokenAuthenticationProvider implements AuthenticationProv
 			}
 		}
 	};
+
+	public get accessTokenProviderInstance(): AccessTokenProvider {
+		return this.accessTokenProvider;
+	}
```
baywet, line 41:
```suggestion
```
Sorry, because of TS syntax I didn't realize that the field was actually public readonly, in effects, we don't need a getter.

baywet (approved):

PR #1578, opened 2025-01-29
fix: implements serialization of enum collections

rkodev:
Partially addresses https://github.com/microsoft/kiota-typescript/issues/1276

- [x] Adds method `writeCollectionOfEnumValue` to the `SerializationWriter` interface and implements it in the `FormSerializationWriter`, `JsonSerializationWriter`, and `MultipartSerializationWriter` classes. 
- [x] Update `getCollectionOfEnumValue` to serialize flaggable enums

PR #1580, opened 2025-01-30
feat: add abstraction.serializationWriter.writeCollectionOfEnumValue 

rkodev:
Adds `abstraction.serializationWriter.writeCollectionOfEnumValue` to be able to compile PR https://github.com/microsoft/kiota-typescript/pull/1578

### DIFF
--- a/packages/serialization/form/src/formSerializationWriter.ts
+++ b/packages/serialization/form/src/formSerializationWriter.ts
@@ -131,6 +131,9 @@ export class FormSerializationWriter implements SerializationWriter {
 			}
 		}
 	};
+	// placeholder for the method
+	// eslint-disable-next-line @typescript-eslint/no-empty-function
+	public writeCollectionOfEnumValue = <T>(_key?: string, _values?: (T | null | undefined)[]): void => {};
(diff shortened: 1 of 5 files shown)


packages/serialization/form/src/formSerializationWriter.ts
```diff
@@ -131,6 +131,9 @@ export class FormSerializationWriter implements SerializationWriter {
 			}
 		}
 	};
+	// placeholder for the method
+	// eslint-disable-next-line @typescript-eslint/no-empty-function
+	public writeCollectionOfEnumValue = <T>(_key?: string, _values?: (T | null | undefined)[]): void => {};
```
baywet, line 136: why do we need placeholder methods here instead of rolling out the implementation directly?

rkodev: This is currently blocking the testing in PR https://github.com/microsoft/kiota-typescript/pull/1578, the serialization interface cant be referenced in tests

rkodev: Going to close this PR, had an issue with testing changes for the unpublished abstraction interface.

Issue #1244, opened 2024-06-21
Singletons are not singletons in Typescript


andreaTP:
In Apicurio Registry we are attempting to use the client SDK generated for Typescript ( [link](https://github.com/Apicurio/apicurio-registry/pull/4762) ).

More specifically, we are following the consolidated pattern of:
- generating the SDK standalone (we will publish it eventually)
- use it as a package from the UI application

Although we started running into issues:
![Screenshot from 2024-06-21 15-46-31](https://github.com/microsoft/kiota-typescript/assets/5792097/5640b823-90e6-4dc4-989a-ec8a384b2843)

After much research and debugging we found out that `ParseNodeFactoryRegistry.defaultInstance` doesn't always refer to the same object, and, depending on: "from which module are you invoking it" you get a different instance of it ( there are multiple resources in SO that refer to this [e.g.](https://stackoverflow.com/questions/55324510/how-to-make-a-singleton-in-typescript-work-across-multiple-modules) ).

### Workaround

Running `kiota generate --serializer none --deserializer none` and registering manually the serializers and deserializers in the end module, i.e.:

```ts
const pnfr = new ParseNodeFactoryRegistry();
const json = new JsonParseNodeFactory();
...
    
pnfr.contentTypeAssociatedFactories.set(json.getValidContentType(), json);
...

new FetchRequestAdapter(new AnonymousAuthenticationProvider(), pnfr, ...);
```

### Solution?

This bug is hard to discover and track down, and it easily compromises the "getting started" experience.
There are a few options on how to solve it, but I'm happy to listen to a proposal from someone more expert than me in TS.

andreaTP:
cc. @EricWittmann

baywet:
Thanks for bringing this up. It's most likely caused because the bundled application actually has two copies of the abstractions, and yes that's a pain. Hard to investigate for people new to kiota and/or to JavaScript.

As part of making the clients truly portable, we're planning to build bundle packages which would do two things:
- refer to abstractions + all the default implementations (except for auth maybe)
- register the serialization providers (through a derived request adapter, not through the singleton registry anymore)
This should solve this issue. We'll still have a singleton for the backing store factory, which could pause similar issues, but we don't have plans for that yet.

andreaTP:
> It's most likely caused because the bundled application actually has two copies of the abstractions

Haven't realized! Why?

> register the serialization providers through a derived request adapter

Do you mean attaching the `defaultInstance` to the adapter?
[...]

PR #1634, opened 2025-03-10
fix:  Removes singleton registries

rkodev:
Resolves https://github.com/microsoft/kiota-typescript/issues/1244


packages/abstractions/src/serialization/kiotaJsonSerializer.ts
```diff
@@ -8,64 +8,72 @@ import { deserialize, deserializeCollection, serialize, serializeCollection, ser
...
 /**
  * Serializes a parsable object into a buffer
+ * @param serializationWriterFactoryRegistry the serialization writer factory registry
  * @param value the value to serialize
  * @param serializationFunction the serialization function for the model type
  * @returns a buffer containing the serialized value
  */
-export function serializeToJson<T extends Parsable>(value: T, serializationFunction: ModelSerializerFunction<T>): ArrayBuffer {
-	return serialize(jsonContentType, value, serializationFunction);
+export function serializeToJson<T extends Parsable>(serializationWriterFactoryRegistry: SerializationWriterFactoryRegistry, value: T, serializationFunction: ModelSerializerFunction<T>): ArrayBuffer {
```
baywet, line 22: the goal of those methods was to make it easy to serialize/deserialize.
Since we need an instance, what do you think about making them instance methods of the `SerializationWriterFactoryRegistry` (and equivalent) classes instead?

packages/http/fetch/src/fetchRequestAdapter.ts
```diff
@@ -356,8 +359,8 @@ export class FetchRequestAdapter implements RequestAdapter {
 		});
 	};
 	public enableBackingStore = (backingStoreFactory?: BackingStoreFactory): void => {
-		this.parseNodeFactory = enableBackingStoreForParseNodeFactory(this.parseNodeFactory);
-		this.serializationWriterFactory = enableBackingStoreForSerializationWriterFactory(this.serializationWriterFactory);
+		this.parseNodeFactory = enableBackingStoreForParseNodeFactory(this.parseNodeFactory as ParseNodeFactoryRegistry, this.parseNodeFactory);
```
baywet, line 362: we should type test instead of using as here.

packages/bundle/src/defaultRequestAdapter.ts
```diff
@@ -24,18 +24,18 @@ export class DefaultRequestAdapter extends FetchRequestAdapter {
...
-	private static setupDefaults() {
-		registerDefaultSerializer(JsonSerializationWriterFactory);
-		registerDefaultSerializer(TextSerializationWriterFactory);
-		registerDefaultSerializer(FormSerializationWriterFactory);
-		registerDefaultSerializer(MultipartSerializationWriterFactory);
-		registerDefaultDeserializer(JsonParseNodeFactory);
-		registerDefaultDeserializer(TextParseNodeFactory);
-		registerDefaultDeserializer(FormParseNodeFactory);
+	private setupDefaults() {
+		registerDefaultSerializer(super.getSerializationWriterFactory() as SerializationWriterFactoryRegistry, JsonSerializationWriterFactory);
```
baywet, line 33: type test instead of as

packages/abstractions/src/apiClientBuilder.ts
```diff
@@ -9,53 +9,58 @@ import { BackingStoreParseNodeFactory, BackingStoreSerializationWriterProxyFacto
 
 /**
  * Registers the default serializer to the registry.
+ * @param serializationWriterFactoryRegistry The registry to which the default serializer will be registered.
  * @param type the class of the factory to be registered.
  */
-export function registerDefaultSerializer(type: new () => SerializationWriterFactory): void {
+export function registerDefaultSerializer(serializationWriterFactoryRegistry: SerializationWriterFactoryRegistry, type: new () => SerializationWriterFactory): void {
```
baywet, line 15: those should probably be instance methods now that we require an instance to be passed

baywet (changes requested):

packages/http/fetch/src/fetchRequestAdapter.ts
rkodev: @baywet, I think we can also move this functions to instance methods as well? I believe all the functions in file `apiClientBuilder.ts` can be moved to instance methods in the registries

packages/abstractions/src/apiClientBuilder.ts
rkodev: Had to revert to functions since they will require [cyclic dependencies](https://github.com/microsoft/kiota-typescript/actions/runs/13830232414/job/38692659039?pr=1634) that are [problematic](https://stackoverflow.com/questions/43176006/typeerror-class-extends-value-undefined-is-not-a-function-or-null)

rkodev: Moving `enableBackingStoreForParseNodeRegistry` to the parseNodeFactoryRegistry will introduce an import dependecy on `BackingStoreParseNodeFactory` which is extends `ParseNodeProxyFactory -> ParseNodeFactory` , since the registry also extends `ParseNodeFactory`, the issues is resulting from 'importing a file that is alsp being extended`

packages/serialization/json/src/browser/jsonParseNodeFactory.ts
```diff
@@ -5,11 +5,17 @@
...
+import { type ParseNode, type ParseNodeFactory, BackingStoreFactory } from "@microsoft/kiota-abstractions";
 
 import { JsonParseNode } from "./../jsonParseNode";
 
 export class JsonParseNodeFactory implements ParseNodeFactory {
+	/**
+	 * Creates an instance of JsonParseNode.
+	 * @param backingStoreFactory - The factory to create backing stores.
+	 */
+	constructor(private readonly backingStoreFactory: BackingStoreFactory) {}
```
baywet, line 17: optional parameter

packages/serialization/text/src/textParseNodeFactory.ts
```diff
@@ -5,11 +5,16 @@
...
+import { BackingStoreFactory, type ParseNode, type ParseNodeFactory } from "@microsoft/kiota-abstractions";
 
 import { TextParseNode } from "./textParseNode";
 
 export class TextParseNodeFactory implements ParseNodeFactory {
+	/**
+	 * Creates an instance of TextParseNode.
+	 * @param backingStoreFactory - The factory to create backing stores.
+	 */
+	constructor(private readonly backingStoreFactory?: BackingStoreFactory) {}
```
baywet, line 17: why ask for the parameter if we're not doing anything with it?

packages/abstractions/src/apiClientBuilder.ts
baywet: right, thanks for articulating this!

packages/serialization/form/src/browser/formParseNodeFactory.ts
```diff
@@ -5,11 +5,16 @@
...
+import type { BackingStoreFactory, ParseNode, ParseNodeFactory } from "@microsoft/kiota-abstractions";
 
 import { FormParseNode } from "./../formParseNode";
 
 export class FormParseNodeFactory implements ParseNodeFactory {
+	/**
+	 * Creates an instance of JsonParseNode.
+	 * @param backingStoreFactory - The factory to create backing stores.
+	 */
+	constructor(private readonly backingStoreFactory: BackingStoreFactory) {}
```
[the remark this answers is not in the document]
baywet:
```suggestion
	constructor(private readonly backingStoreFactory?: BackingStoreFactory) {}
```

baywet:
@rkodev a couple of things before we can merge this:
- I believe the reason why you had added the parameter for the backing store for text, was to have a consistent API surface across different formats? If so, we might actually want to put this back as this impact code generation. Sorry for not "getting it" earlier.
- Have you started the code generation changes?
- There are still two comments open, if you could simply provide answers there so we can burn them down.

packages/bundle/src/defaultRequestAdapter.ts
rkodev: This should have been resolved with the refactor

packages/http/fetch/src/fetchRequestAdapter.ts
rkodev: The error should be resolved, I'm currently type testing

baywet (approved):

If you could start the generation PR before we merge this one?
Also, can you please make sure you squash merge with a conventional commit?

Issue #1652, opened 2025-03-19
[http/fetch] Multiple `span.end()` calls


yasaichi:
I am wondering if multiple `span.end()` calls are intented in the following code of `fetchRequestAdapter.ts`:

- `startTracingSpan`: https://github.com/microsoft/kiota-typescript/blob/1dd543235361c045680e6171f0c5db6426c71cb5/packages/http/fetch/src/fetchRequestAdapter.ts#L168
- `send` or other `send*` methods: https://github.com/microsoft/kiota-typescript/blob/1dd543235361c045680e6171f0c5db6426c71cb5/packages/http/fetch/src/fetchRequestAdapter.ts#L203

I do not think it is necessary to call `span.end()` in the `finally` block of the `send` method because it is called in the one of `startTracingSpan`. In fact, this leads to the following warning messages of an application I have been working on:

```
Cannot execute the operation on ended Span {traceId: 1eda2ce930a1e6476498849531bca2cb, spanId: 3ec08dd4f1b3a1bf} Error: Operation attempted on ended Span {traceId: 1eda2ce930a1e6476498849531bca2cb, spanId: 3ec08dd4f1b3a1bf}
    at SpanImpl._isSpanEnded (file:///tmp/deno-compile-server/node_modules/.pnpm/@opentelemetry+sdk-trace-base@2.0.0_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-trace-base/build/src/Span.js:266:27)
    at SpanImpl.end (file:///tmp/deno-compile-server/node_modules/.pnpm/@opentelemetry+sdk-trace-base@2.0.0_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-trace-base/build/src/Span.js:178:18)
    at file:///tmp/deno-compile-server/node_modules/.pnpm/@microsoft+kiota-http-fetchlibrary@1.0.0-preview.72/node_modules/@microsoft/kiota-http-fetchlibrary/dist/es/src/fetchRequestAdapter.js:166:26
    at eventLoopTick (ext:core/01_core.js:177:7)

...
send - {+baseurl}/foo/bar - You can only call end() on a span once.
```

PR #1654, opened 2025-03-19
Fix `Error: Operation attempted on ended Span`

yasaichi:
Fixes https://github.com/microsoft/kiota-typescript/issues/1652


yasaichi:

baywet (approved):
Thanks for the contribution!

PR #1657, opened 2025-03-19
Fix json parser ArrayBuffer memory leak

### LINKED ISSUE #1653: ArrayBuffer possible memory leaking
We have just concluded the investigation into our app leaking memory after a recent Kiota update. Basically, what happened is that Kiota started generating "proper type" if the format of a field is specified as "bytes." Before the update, it was of string type and everything worked correctly. We did deserialization on our side as the field was sent properly as a base64 encoded byte array in string.

However, after a recent update which possibly added the generation for that format, we started encountering major memory leaks. We think this might be because of integration with React (though we're not 100% certain about that). The issue appears to be that `Buffer.from(strValue, "base64")` allocates a buffer which can then be reused by Node after certain conditions are met, possibly if you are not working with the buffer instantly at that moment.

We have been getting buffers which contain random console/file output in them instead of the proper values. So we tried creating our own JsonParseNode by taking basically the code of the original and just editing the getByteArrayValue method like this:

```typescript
public getByteArrayValue(): ArrayBuffer | undefined {
    const strValue = this.getStringValue();
    if (strValue && strValue.length > 0) {
      if (inNodeEnv()) {
        // In Node.js, create a new buffer from the decoded base64 data
        // This ensures we don't get unexpected data from buffer reuse
        const buffer = Buffer.from(strValue, "base64");
        
        // Create a completely new Uint8Array with only our data
        const cleanBuffer = new Uint8Array(buffer.length);
        for (let i = 0; i < buffer.length; i++) {
          cleanBuffer[i] = buffer[i];
        }

        return cleanBuffer.buffer;
      } else {
        // In browser, use the built-in base64 decoder
        try {
          const binaryStr = atob(strValue);
          const bytes = new Uint8Array(binaryStr.length);
          for (let i = 0; i < binaryStr.length; i++) {
            bytes[i] = binaryStr.charCodeAt(i);
          }
          return bytes.buffer;
        } catch (e) {
          console.error("Failed to decode base64 string:", e);
          return undefined;
        }
      }
    }
    return undefined;
  }
```

This is more of a discussion than a bug report, but at the same time it feels like this might be a bug. We would like to hear the team's thoughts on this.

zZHorizonZz:
This PR fixes ArrayBuffer memor leak as described in https://github.com/microsoft/kiota-typescript/issues/1653

Closes #1653

### DIFF
--- a/packages/serialization/json/src/jsonParseNode.ts
+++ b/packages/serialization/json/src/jsonParseNode.ts
@@ -48,7 +48,19 @@ export class JsonParseNode implements ParseNode {
 	public getByteArrayValue(): ArrayBuffer | undefined {
 		const strValue = this.getStringValue();
 		if (strValue && strValue.length > 0) {
-			return inNodeEnv() ? Buffer.from(strValue, "base64").buffer : new TextEncoder().encode(strValue);
+			/**
+			 * Node.js Buffer objects created via Buffer.from() use a shared memory pool
+			 * and may be subject to reuse/recycling, which can lead to unexpected behavior.
+			 * 
+			 * For consistent behavior across environments:
+			 * - In Node: We convert the base64 string to a Buffer first, then create a new
+			 *   Uint8Array from it to ensure we have a stable, independent copy
+			 * - In browsers: We convert the string directly using TextEncoder
+			 * 
+			 * TODO: Consider standardizing on a cross-platform UInt8Array.fromBase64 (after the API is stabilized across browsers) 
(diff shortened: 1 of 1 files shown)


baywet (dismissed):
Thanks for the contribution!

zZHorizonZz:

baywet (approved):

zZHorizonZz:
These seem like formatting errors. I didn't really run format tasks as I did edits through GitHub's web vs code.

baywet (approved):

PR #1798, opened 2025-07-30
Update apiClientProxifier.ts

### LINKED ISSUE #1725: couldn't find navigation property then data
Looks like the apiClientProxifier blows up when an instance is returned from an async function.  I am loading some server-side configuration options before generating the client and wanted to return the client instance in a single pass.

```
export async function getApiClient() : Promise<ApiClient> {
  const ibOptions = await fetch('/api/auth/client-config').then(r => r.json());
  const credential = new InteractiveBrowserCredential(ibOptions);  
  const authProvider = new AzureIdentityAuthenticationProvider(credential, ['api://<api-client-id>/scopename']);
  const adapter = new FetchRequestAdapter(authProvider);
  adapter.baseUrl = location.origin;
  
  var client = createApiClient(adapter);
  return client;
}
```

When I call this method, I receive the error message `couldn't find navigation property then data`.

It looks like what is needed, is inside `apiClientProxifier` is a special name check for `then` that simply returns `undefined` in order to satisfy the Promise resolution logic.

tracker1:
fix: Allow API Client Proxy to be returned from an async function. 

Closes #1725

### DIFF
--- a/packages/abstractions/src/apiClientProxifier.ts
+++ b/packages/abstractions/src/apiClientProxifier.ts
@@ -124,6 +124,9 @@ export const apiClientProxifier = <T extends object>(requestAdapter: RequestAdap
 					return apiClientProxifier(requestAdapter, getPathParameters(rawUrl), navigationMetadata, requestsMetadata);
 				};
 			}
+      if (name === "then") {
+        return undefined;
+      }
 			if (requestsMetadata) {
 				const metadataKey = getRequestMethod(name);
 				if (metadataKey) {


packages/abstractions/src/apiClientProxifier.ts
```diff
@@ -124,6 +124,9 @@ export const apiClientProxifier = <T extends object>(requestAdapter: RequestAdap
 					return apiClientProxifier(requestAdapter, getPathParameters(rawUrl), navigationMetadata, requestsMetadata);
 				};
 			}
+      if (name === "then") {
```
baywet, line 127: It'd probably be safer to move this block before the error is thrown.
My worry is that if somebody has an API segment called "then" it will derail them.
At this other location, we'll know we didn't find any navigation metadata which makes for a more specific condition.
What do you think?

baywet (changes requested):
Thanks for the contribution!

In addition to the requested changes, can you please add a unit test to prevent any regression?

packages/abstractions/src/apiClientProxifier.ts
```diff
@@ -204,6 +204,10 @@ export const apiClientProxifier = <T extends object>(requestAdapter: RequestAdap
 						return apiClientProxifier(requestAdapter, downWardPathParameters, navigationCandidate.navigationMetadata, navigationCandidate.requestsMetadata);
 					};
 				}
+        if (name === "then") {
```
baywet, line 207: prettier will most likely complain about the indent. You can fix it by running `npm run prettier:write`

ofekR123:
Hi! Is there any update regarding this bug?

JustinGrote:
Here's a pnpm hotpatch for preview-97 that implements the PR. Thanks @tracker1!
```diff
diff --git a/dist/es/src/apiClientProxifier.js b/dist/es/src/apiClientProxifier.js
index 748655aa7579398e0b3ef061f653b8550e2e26ed..dcd4c00fd114ebeaf5f423a5cd2bc0c8b77d406e 100644
--- a/dist/es/src/apiClientProxifier.js
+++ b/dist/es/src/apiClientProxifier.js
@@ -208,6 +208,10 @@ export const apiClientProxifier = (requestAdapter, pathParameters, navigationMet
                         return apiClientProxifier(requestAdapter, downWardPathParameters, navigationCandidate.navigationMetadata, navigationCandidate.requestsMetadata);
                     };
                 }
+                if (name === "then") {
+                    // handle returning proxy from an async method
+                    return undefined
+                }
                 throw new Error(`couldn't find navigation property ${name} data: ${JSON.stringify(navigationMetadata)}`);
             }
         },

```

baywet:
@JustinGrote Thank you for using kiota and for reaching out.
I believe this PR is abandoned as it is, and I suggest you open a new one to fast track things. Don't hesitate to salvage whatever you can from here!

ofekR123:
@JustinGrote did this reach the main branch? I couldn't find it not there and not in the kiota bundle 97 perview tag?

JustinGrote:
> @JustinGrote did this reach the main branch? I couldn't find it not there and not in the kiota bundle 97 perview tag?

I didn't create a PR, I provided a patch that I am using via pnpm. I am not interested in going through a lengthy PR process but you can feel free to take my changes as PR them.

baywet:
CC @adrian05-ms @MIchaelMainer @gavinbarron this needs attention

PR #2039, opened 2026-06-06
fix(serialization): pass primitive collection type

### LINKED ISSUE #2000: getCollectionOfPrimitiveValues should accept a type as a parameter
While working on #1964 and #1963 it became apparent that the getCollectionOfPrimitiveValues method cannot work as expected since the underlying "data" is always string for form and text.
It does work for JSON since the JSON decoding does some of the work of type mapping for us.

This method should accept a typeof as a parameter (we need to update the interface and the code generation as well) since the generic type parameter is erased when transpiling.
Then the implementation can use that parameter value to return the correct type information.

nanookclaw:
Closes #2000.

This updates `ParseNode.getCollectionOfPrimitiveValues` to accept an optional runtime primitive type, while keeping the no-argument call path compatible. The form parse node now uses that runtime type to deserialize comma-separated primitive collections as strings, numbers, booleans, dates, `DateOnly`, `TimeOnly`, or `Duration`. JSON and text parse nodes accept the same parameter so the interface stays consistent; JSON keeps returning the already-decoded values.

The fetch request adapter now passes the requested primitive response type into collection deserialization, including `Duration`, `DateOnly`, and `TimeOnly` collection response types.

Validation run locally:
- `npm --workspace @microsoft/kiota-abstractions run build`
- `npm --workspace @microsoft/kiota-serialization-form run test:node`
- `npm --workspace @microsoft/kiota-serialization-json run test:node`
- `npm --workspace @microsoft/kiota-serialization-form run build`
- `npm --workspace @microsoft/kiota-serialization-json run build`
- `npm --workspace @microsoft/kiota-serialization-text run build`
- `npm --workspace @microsoft/kiota-http-fetchlibrary run build`
- `npm run prettier:check`
- `npx eslint --quiet packages/abstractions/src/serialization/parseNode.ts packages/http/fetch/src/fetchRequestAdapter.ts packages/serialization/form/src/formParseNode.ts packages/serialization/json/src/jsonParseNode.ts packages/serialization/text/src/textParseNode.ts`
- `git diff --check`

I also ran `npm run lint`; it still reports an existing unrelated error in `packages/bundle/src/defaultRequestAdapter.ts`.

### DIFF
--- a/packages/abstractions/src/requestAdapter.ts
+++ b/packages/abstractions/src/requestAdapter.ts
@@ -110,8 +110,10 @@ export type PrimitiveTypesForDeserializationType = string | number | boolean | D
 
 export type PrimitiveTypesForDeserialization = "string" | "number" | "boolean" | "Date" | "DateOnly" | "TimeOnly" | "Duration" | "ArrayBuffer";
 
+/** Primitive value types supported for collection response deserialization. */
 export type PrimitiveTypesForDeserializationTypeForCollection = Exclude<PrimitiveTypesForDeserializationType, ArrayBuffer>;
 
+/** Runtime primitive type names supported for collection response deserialization. */
 export type PrimitiveTypesForDeserializationForCollection = Exclude<PrimitiveTypesForDeserialization, "ArrayBuffer">;
 
 export type SendMethods = Exclude<keyof RequestAdapter, "enableBackingStore" | "getSerializationWriterFactory" | "convertToNativeRequest" | "baseUrl">;
--- a/packages/abstractions/src/requestAdapter.ts
+++ b/packages/abstractions/src/requestAdapter.ts
@@ -51,7 +51,7 @@ export interface RequestAdapter {
 	 * @param errorMappings the error factories mapping to use in case of a failed request.
 	 * @returns a {@link Promise} with the deserialized response model collection.
 	 */
-	sendCollectionOfPrimitive<ResponseType extends Exclude<PrimitiveTypesForDeserializationType, ArrayBuffer>>(requestInfo: RequestInformation, responseType: Exclude<PrimitiveTypesForDeserialization, "ArrayBuffer">, errorMappings: ErrorMappings | undefined): Promise<ResponseType[] | undefined>;
+	sendCollectionOfPrimitive<ResponseType extends PrimitiveTypesForDeserializationTypeForCollection>(requestInfo: RequestInformation, responseType: PrimitiveTypesForDeserializationForCollection, errorMappings: ErrorMappings | undefined): Promise<ResponseType[] | undefined>;
 	/**
 	 * Executes the HTTP request specified by the given RequestInformation and returns the deserialized primitive response model.
 	 * @param requestInfo the request info to execute.
@@ -109,5 +109,9 @@ export interface ErrorMappings {
 export type PrimitiveTypesForDeserializationType = string | number | boolean | Date | DateOnly | TimeOnly | Duration | ArrayBuffer;
 
 export type PrimitiveTypesForDeserialization = "string" | "number" | "boolean" | "Date" | "DateOnly" | "TimeOnly" | "Duration" | "ArrayBuffer";
+
+export type PrimitiveTypesForDeserializationTypeForCollection = Exclude<PrimitiveTypesForDeserializationType, ArrayBuffer>;
+
+export type PrimitiveTypesForDeserializationForCollection = Exclude<PrimitiveTypesForDeserialization, "ArrayBuffer">;
 
 export type SendMethods = Exclude<keyof RequestAdapter, "enableBackingStore" | "getSerializationWriterFactory" | "convertToNativeRequest" | "baseUrl">;
--- a/packages/abstractions/src/serialization/parseNode.ts
+++ b/packages/abstractions/src/serialization/parseNode.ts
@@ -8,6 +8,7 @@ import type { Guid } from "../utils/guidUtils";
 
 import type { DateOnly } from "../dateOnly";
 import type { Duration } from "../duration";
+import type { PrimitiveTypesForDeserialization } from "../requestAdapter";
 import type { TimeOnly } from "../timeOnly";
 import type { Parsable } from "./parsable";
 import type { ParsableFactory } from "./parsableFactory";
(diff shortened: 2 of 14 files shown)


packages/abstractions/src/serialization/parseNode.ts
```diff
@@ -64,9 +65,10 @@ export interface ParseNode {
 	getTimeOnlyValue(): TimeOnly | undefined;
 	/**
 	 * Gets the collection of primitive values of the node.
+	 * @param primitiveType the runtime primitive type to deserialize collection items into.
 	 * @returns the collection of primitive values of the node.
 	 */
-	getCollectionOfPrimitiveValues<T>(): T[] | undefined;
+	getCollectionOfPrimitiveValues<T>(primitiveType?: Exclude<PrimitiveTypesForDeserialization, "ArrayBuffer">): T[] | undefined;
```
baywet: could we define a type alias to avoid duplicating the exclude syntax all over the place?

packages/serialization/json/src/jsonParseNode.ts
baywet: The code here needs to be updated to rely on the passed parameter instead.

packages/serialization/text/src/textParseNode.ts
```diff
@@ -58,7 +58,7 @@ export class TextParseNode implements ParseNode {
 	public getDateOnlyValue = () => DateOnly.parse(this.getStringValue());
 	public getTimeOnlyValue = () => TimeOnly.parse(this.getStringValue());
 	public getDurationValue = () => Duration.parse(this.getStringValue());
-	public getCollectionOfPrimitiveValues = <T>(): T[] | undefined => {
+	public getCollectionOfPrimitiveValues = <T>(_primitiveType?: Exclude<PrimitiveTypesForDeserialization, "ArrayBuffer">): T[] | undefined => {
```
baywet: implementation needs updated to rely on the parameter instead.

baywet (changes requested):
Thanks for the contribution!

In addition to the comment I've left here, can you also stage the corresponding generation pull request please?

nanookclaw:
Addressed the review feedback in `5c9ebc31`:

- added collection-specific primitive aliases in abstractions and used them instead of repeated `Exclude<...ArrayBuffer>` expressions
- updated JSON collection primitive deserialization to honor the provided runtime primitive type for `Date`, `DateOnly`, `TimeOnly`, and `Duration`
- kept the no-parameter JSON behavior backward-compatible

Checks run locally:
- `npm run build:esm -w @microsoft/kiota-abstractions`
- `npm run build:esm -w @microsoft/kiota-serialization-json`
- `npm run build:esm -w @microsoft/kiota-http-fetchlibrary`
- `npm run test:node -w @microsoft/kiota-serialization-json`
- `npx prettier --check ...`

I’ll stage the corresponding generation PR next.

packages/abstractions/src/requestAdapter.ts
```diff
@@ -110,4 +110,8 @@ export type PrimitiveTypesForDeserializationType = string | number | boolean | D
 
 export type PrimitiveTypesForDeserialization = "string" | "number" | "boolean" | "Date" | "DateOnly" | "TimeOnly" | "Duration" | "ArrayBuffer";
 
+export type PrimitiveTypesForDeserializationTypeForCollection = Exclude<PrimitiveTypesForDeserializationType, ArrayBuffer>;
```
baywet, line 114: why do we need both here?

baywet (commented):

nanookclaw:
Pushed `cbba8f0d` to address the Node 24 lint failure from CI.

The only change marks `JsonParseNode.getPrimitiveValue` as `private readonly`, matching the `@typescript-eslint/prefer-readonly` diagnostic.

Local verification:
- `git diff --check`
- `npx eslint --quiet packages/serialization/json/src/jsonParseNode.ts`
- `npm run build:esm -w @microsoft/kiota-serialization-json`

packages/serialization/json/src/jsonParseNode.ts
baywet: Great first step refactoring into a method. I think we can remove `const typeOfX = typeof x;` and all the if/else block that relies on it then.

baywet (commented):
Making progress. Thanks for the updates. I think a few comments still need to be addressed. Let me know if you have any additional comments or questions.

packages/serialization/json/src/jsonParseNode.ts
nanookclaw: Addressed in d699888: json primitive collections now use the passed primitiveType via getPrimitiveValue, defaulting to string for existing generated string-array callers. The old typeof inference block is removed.

packages/serialization/text/src/textParseNode.ts
nanookclaw: Addressed in d699888: text primitive collections now rely on the primitiveType parameter when splitting/deserializing values, matching the runtime-type path used by the request adapter.

packages/abstractions/src/requestAdapter.ts
nanookclaw: Addressed in d699888 by clarifying the two exported aliases in place: TypeForCollection is the compile-time value union for the generic response type, while ForCollection is the runtime string discriminator passed to parse nodes. Both intentionally exclude ArrayBuffer for collection responses.

packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -55,35 +55,11 @@ export class JsonParseNode implements ParseNode {
 	public getDateOnlyValue = () => this.getDateOnlyValueFromRaw(this._jsonNode);
 	public getTimeOnlyValue = () => this.getTimeOnlyValueFromRaw(this._jsonNode);
 	public getDurationValue = () => this.getDurationValueFromRaw(this._jsonNode);
-	public getCollectionOfPrimitiveValues = <T>(primitiveType?: PrimitiveTypesForDeserializationForCollection): T[] | undefined => {
+	public getCollectionOfPrimitiveValues = <T>(primitiveType: PrimitiveTypesForDeserializationForCollection = "string"): T[] | undefined => {
```
baywet: let's not default the value here, but make it mandatory instead

nanookclaw: Addressed in ee567f8: `primitiveType` is now mandatory on `ParseNode.getCollectionOfPrimitiveValues` and the JSON/form/text parse nodes no longer default to `"string"`. Existing string collection callers/tests pass `"string"` explicitly.

Verification:
- `npm run build --workspace @microsoft/kiota-abstractions`
- `npm run build --workspace @microsoft/kiota-serialization-json`
- `npm run build --workspace @microsoft/kiota-serialization-form`
- `npm run build --workspace @microsoft/kiota-serialization-text`
- JSON/form/text node tests
- `npx tsc -p packages/test/tsconfig.json --noEmit`
- `git diff --check`

baywet:
@nanookclaw thanks for addressing all the comments. I think we're good to go on this side. Can you please start the related generation pull request so the correct code gets emitted?
Let me know if you have any additional comments or questions.

nanookclaw:
Opened the generator-side follow-up at microsoft/kiota#7800.

It updates TypeScript generation for primitive collection fields to emit the runtime type argument expected by this PR, e.g. `n.getCollectionOfPrimitiveValues<string>("string")`, with coverage for the supported primitive tokens. Local focused generator tests passed in a .NET 10 SDK container.

baywet (approved):

PR #2086, opened 2026-07-23
Support map-style query parameters (type: object + additionalProperties)

bkoelman:
Contributes to https://github.com/microsoft/kiota/issues/3800.

Extend RequestInformation to handle plain-object (dictionary) query parameter values so that RFC 6570 {?param*} templates expand them into individual key=value pairs, matching OAS 4.8.12.1 behaviour for exploded form-style query parameters.

Three gaps are closed:
- queryParameters now accepts Record<string, string | null> entries.
- setQueryStringParametersFromRawObject stores plain-object values instead of silently ignoring them.
- normalizeValue routes objects through the new sanitizeMapValue helper, which strips null/undefined entries before they reach StdUriTemplate (null values would otherwise cause convertNativeTypes to throw because typeof null === "object").

Null-entry semantics follow RFC 6570 para 2.3: an undefined/null value produces no output; use "" to emit ?key= (empty value present).

### DIFF
--- a/packages/abstractions/src/requestInformation.ts
+++ b/packages/abstractions/src/requestInformation.ts
@@ -65,12 +65,12 @@ export class RequestInformation implements RequestInformationSetContent {
 			const data = {} as Record<string, unknown>;
 			for (const key in this.queryParameters) {
 				if (this.queryParameters[key] !== null && this.queryParameters[key] !== undefined) {
-					data[key] = this.normalizeValue(this.queryParameters[key]);
+					data[key] = RequestInformation.normalizeValue(this.queryParameters[key]);
 				}
 			}
 			for (const key in this.pathParameters) {
 				if (this.pathParameters[key] !== null && this.pathParameters[key] !== undefined) {
-					data[key] = this.normalizeValue(this.pathParameters[key]);
+					data[key] = RequestInformation.normalizeValue(this.pathParameters[key]);
 				}
 			}
 			return StdUriTemplate.expand(this.urlTemplate, data);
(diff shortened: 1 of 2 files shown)


packages/abstractions/src/requestInformation.ts
```diff
@@ -249,8 +251,27 @@ export class RequestInformation implements RequestInformationSetContent {
...
 		return value;
 	}
+
+	/**
+	 * Pre-processes a plain-object dictionary for StdUriTemplate expansion.
+	 * Null/undefined values are omitted: RFC 6570 §2.3 treats "undefined" variables
+	 * as absent (no output). Use "" to send ?key= (empty value).
+	 */
+	private static sanitizeMapValue(map: Record<string, unknown>): Record<string, string> {
+		const result: Record<string, string> = {};
```
baywet: same as dotnet and java, can we use record of string object here to avoid the string normalization?

baywet (changes requested):

bkoelman:
pushed another commit

packages/abstractions/src/requestInformation.ts
```diff
@@ -89,7 +89,7 @@ export class RequestInformation implements RequestInformationSetContent {
 	/** The Request Body. */
 	public content?: ArrayBuffer;
 	/** The Query Parameters of the request. */
-	public queryParameters: Record<string, string | number | boolean | string[] | number[] | undefined> = createRecordWithCaseInsensitiveKeys<string | number | boolean | string[] | number[] | undefined>();
+	public queryParameters: Record<string, string | number | boolean | string[] | number[] | Record<string, unknown> | undefined> = createRecordWithCaseInsensitiveKeys<string | number | boolean | string[] | number[] | Record<string, unknown> | undefined>();
```
baywet, line 92: unknown casts a wide net. Could we constrain it to primitive types?

PR #2103, opened 2026-08-18
ci: removes the private feed check from required checks

baywet:
unblocks #2102 For security reasons, ADO pipelines don't run on PR from forks. So we can't have it as a check, otherwise it'll block any PR from forks


gavinbarron (approved):
Thanks for catching the fork case,

PR #2132, opened 2026-09-12
fix(abstractions): accept defined falsy multipart content

### LINKED ISSUE #2032: MultipartBody.addOrReplacePart throws unexpectedly due to overly broad falsy check on content
`MultipartBody.addOrReplacePart()` throws `Error: "content cannot be undefined"` for any falsy `content` value due to the guard using `if (!content)`. This incorrectly rejects legitimate values including:

- `""` (empty string)
- `0` (number)
- `0n` (BigInt)
- `false` (boolean)
- `null`

The check should be narrowed to only guard against `undefined`. This would also make it consistent with the error message it throws. The fix in [packages/abstractions/src/multipartBody.ts](https://github.com/microsoft/kiota-typescript/blob/049412c9/packages/abstractions/src/multipartBody.ts#L37):

```ts
// current
if (!content) throw new Error("content cannot be undefined");

// should be
if (content === undefined) throw new Error("content cannot be undefined");
```

**How to reproduce**

```ts
const body = new MultipartBody();
body.addOrReplacePart("emptyString", "application/json", ""); // throws
body.addOrReplacePart("zero", "application/json", 0); // throws
body.addOrReplacePart("bigIntZero", "application/json", 0n); //throws
body.addOrReplacePart("false", "application/json", false); // throws
body.addOrReplacePart("null", "application/json", null); // throws
```
[Link to StackBlitz reproduction](https://stackblitz.com/edit/vitejs-vite-x5cnsff3?file=src%2Fmain.ts)

**Expected behavior**

All five parts are added successfully.

**Actual behavior**

All five throw `Error: content cannot be undefined`.

**Worth Noting**
- [Checks against partName and contentType](https://github.com/microsoft/kiota-typescript/blob/049412c9/packages/abstractions/src/multipartBody.ts#L33-L36) use the same falsy check. Falsy values in those cases are probably legitimately invalid, but the error wording is still slightly inaccurate.
- There is [a test case titled "implements defensive programming"](https://github.com/microsoft/kiota-typescript/blob/049412c9/packages/abstractions/test/common/multipartBody.ts#L17-L17) that asserts an empty string content throws an error.

rksharma-owg:
`MultipartBody.addOrReplacePart()` rejects empty strings, zero, zero bigint, false, and null with “content cannot be undefined.” Its truthiness guard rejects all falsy values rather than only missing content.

Check `content === undefined` and clarify the parameter documentation. Regression tests cover adding and replacing each reported value, preserving an existing part after a rejected undefined replacement, and serializing an empty string alone or between other fields with correct multipart delimiters.

This changes admission of defined content. It does not add scalar serialization support: direct number, bigint, boolean, and null parts remain subject to the existing multipart serializer's type restrictions. Existing valid string/binary/model serialization and part-name/content-type validation retain their behavior.

Validation on Node 22.23.1:

- Before the fix, 10 abstraction regression cases and both empty-string wire-format cases failed.
- `npm ci`: restored the unchanged lockfile.
- `npm run build`, `npm run build:integrated`, `npm run lint:eslint:loud`, and `git diff --check`: passed. Build includes the repository-wide Prettier check.
- `npm test`: both repository scripts passed, each reporting 394 passed and one existing skipped test. The current `test:browser` script does not enable browser mode, so these are not claimed as two distinct browser environments.
- Actual headless Chrome, using an external Vitest configuration with `@vitest/browser-webdriverio@5.0.0`: both affected packages' common suites passed, 111 tests and one existing skip. No browser-tool dependency or configuration change is included in this PR.
- `npx lerna run test:coverage --scope @microsoft/kiota-abstractions --scope @microsoft/kiota-serialization-multipart`: passed.

The full supported Node/platform matrix was not run locally. No dependency or generated-file changes are included.

Closes #2032.

Prepared with AI assistance; reproduction and validation ran locally.

### DIFF
--- a/packages/abstractions/src/multipartBody.ts
+++ b/packages/abstractions/src/multipartBody.ts
@@ -25,7 +25,7 @@ export class MultipartBody implements Parsable {
 	 * Adds or replaces a part with the given name, content type and content.
 	 * @param partName the name of the part to add or replace.
 	 * @param partContentType the content type of the part to add or replace.
-	 * @param content the content of the part to add or replace.
+	 * @param content the content of the part to add or replace. Must not be undefined.
 	 * @param serializationCallback the serialization callback to use when serializing the part.
 	 * @param fileName the name of the file associated with this part.
 	 */
@@ -34,7 +34,7 @@ export class MultipartBody implements Parsable {
 		if (!partContentType) {
 			throw new Error("partContentType cannot be undefined");
 		}
-		if (!content) throw new Error("content cannot be undefined");
+		if (content === undefined) throw new Error("content cannot be undefined");
 		const normalizePartName = this.normalizePartName(partName);
 		this._parts[normalizePartName] = {
 			contentType: partContentType,
--- a/packages/abstractions/test/common/multipartBody.ts
+++ b/packages/abstractions/test/common/multipartBody.ts
@@ -14,7 +14,7 @@ describe("multipartBody", () => {
 		const mpBody = new MultipartBody();
 		assert.throws(() => mpBody.addOrReplacePart("", "application/json", "test"), Error, "partName cannot be undefined");
 		assert.throws(() => mpBody.addOrReplacePart("test", "", "test"), Error, "partContentType cannot be undefined");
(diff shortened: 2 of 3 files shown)


baywet (approved):
Thank you for the contribution!

PR #2135, opened 2026-09-14
fix(serialization): support no-argument getCollectionOfPrimitiveValues

### LINKED ISSUE #2061: [Regression] preview.103 getCollectionOfPrimitiveValues() with no argument throws, breaking clients generated against the previous signature (e.g. @microsoft/msgraph-sdk)
### Description

`@microsoft/kiota-serialization-json@1.0.0-preview.103` (and `-form`) changed `getCollectionOfPrimitiveValues` to **require** a runtime `primitiveType` argument (#2039). This breaks any client that was generated against the previous no-argument signature, because the old generated deserializers call `n.getCollectionOfPrimitiveValues()` with no argument.

The currently published `@microsoft/msgraph-sdk@1.0.0-preview.83` (latest on npm) is such a client. With kiota serialization bumped to `preview.103`, deserializing any `subscribedSku` (which has a `subscriptionIds` string collection) throws:

```
encountered an unsupported type during deserialization undefined
```

### Root cause

In `preview.103`, `getCollectionOfPrimitiveValues` forwards to `getPrimitiveValue(value, primitiveType)`:

```ts
private readonly getPrimitiveValue = <T>(value: unknown, primitiveType: PrimitiveTypesForDeserializationForCollection): T => {
  if (value === null) return null as T;
  switch (primitiveType) {
    case "string": ...
    case "number": ...
    // ...
    default:
      throw new Error(`encountered an unsupported type during deserialization ${primitiveType as string}`);
  }
};
```

When an older generated client calls `getCollectionOfPrimitiveValues()` without the new argument, `primitiveType` is `undefined`, falls through to `default`, and throws. The trailing `undefined` in the message is the missing `primitiveType`, not the data.

In `preview.102` the method inspected each element's own `typeof` and required no type argument, so the same payload deserialized fine.

Stack trace (from `msgraph-sdk@preview.83`):

```
this.getPrimitiveValue (@microsoft/kiota-serialization-json@1.0.0-preview.103/.../jsonParseNode.ts:93)
this.getCollectionOfPrimitiveValues (.../jsonParseNode.ts:62)
<object>.subscriptionIds (@microsoft/msgraph-sdk@1.0.0-preview.83/models/index.ts)
this.assignFieldValues (.../jsonParseNode.ts:166)
this.getObjectValue (@microsoft/msgraph-sdk-core/.../BatchRequestContent.ts)
```

The relevant generated deserializer in `msgraph-sdk@preview.83` is:

```js
"subscriptionIds": n => { subscribedSku.subscriptionIds = n.getCollectionOfPrimitiveValues() }
```

### Reproduction

1. Install `@microsoft/msgraph-sdk@1.0.0-preview.83` together with `@microsoft/kiota-bundle@1.0.0-preview.103` (which pulls `kiota-serialization-json@1.0.0-preview.103`).
2. Call `graphServiceClient.subscribedSkus.get()` against a tenant.
3. Deserialization of `subscriptionIds` throws `encountered an unsupported type during deserialization undefined`.

Minimal equivalent:

```ts
import { JsonParseNode } from "@microsoft/kiota-serialization-json"; // preview.103
const node = new JsonParseNode(["a", "b"]);
node.getCollectionOfPrimitiveValues(); // no arg -> throws (worked on preview.102)
```

### Expected

`getCollectionOfPrimitiveValues()` called without `primitiveType` should remain backward-compatible (fall back to the previous element-`typeof` behavior) rather than throwing, OR the change should be released as an explicit breaking change requiring all SDKs to be regenerated first.

PR #2039's description states the no-argument call path was meant to stay compatible ("accept an **optional** runtime primitive type, while keeping the no-argument call path compatible") — the `default: throw` in the JSON path breaks that guarantee.

### Environment

- `@microsoft/kiota-serialization-json` / `-form`: `1.0.0-preview.103`
- `@microsoft/msgraph-sdk`: `1.0.0-preview.83` (latest published)
- node: v24

### Related

- #2039 (the change)
- #2000 (the issue it closed)
- Generator follow-up: microsoft/kiota#7800

rksharma-owg:
Closes #2061.

### Problem
PR #2039 added a runtime `primitiveType: PrimitiveTypesForDeserializationForCollection` argument to `ParseNode.getCollectionOfPrimitiveValues`. However:
1. In `packages/abstractions/src/serialization/parseNode.ts`, `primitiveType` was typed as required rather than optional (`primitiveType?: PrimitiveTypesForDeserializationForCollection`).
2. In `JsonParseNode.ts`, `getPrimitiveValue` dispatched via `switch (primitiveType)`. When called with no argument (or `undefined`), it hit `default` and threw `encountered an unsupported type during deserialization undefined`. This broke existing generated SDK clients (such as `@microsoft/msgraph-sdk@1.0.0-preview.83` and other consumers) whose deserializers call `node.getCollectionOfPrimitiveValues()` without arguments.
3. In `FormParseNode.ts` and `TextParseNode.ts`, calling without arguments likewise threw in the `switch (primitiveType)` statement.

### Solution
- Make `primitiveType?: PrimitiveTypesForDeserializationForCollection` optional on `ParseNode.getCollectionOfPrimitiveValues`.
- In `JsonParseNode.getPrimitiveValue`: when `primitiveType === undefined`, fall back to inspecting element runtime types (`boolean`, `string`, `number`, `Date`, `DateOnly`, `TimeOnly`, `Duration`) and preserving `null` values, restoring exact backward compatibility with preview.102.
- In `FormParseNode.getCollectionOfPrimitiveValues`: handle `case undefined:` alongside `case "string":`, preserving default string deserialization.
- In `TextParseNode.getCollectionOfPrimitiveValues`: handle `case undefined:` alongside `case "string":`, preserving default string deserialization.
- In `MockParseNode`: make `_primitiveType?: PrimitiveTypesForDeserializationForCollection` optional to match the interface.
- Add regression tests in `JsonParseNode`, `FormParseNode`, and `TextParseNode` verifying that no-argument calls succeed without throwing.

### Validation
- Unit tests:
  - `npm --workspace @microsoft/kiota-abstractions run build` (passed)
  - `npm --workspace @microsoft/kiota-serialization-json run test:node` (30/30 passed)
  - `npm --workspace @microsoft/kiota-serialization-form run test:node` (13/13 passed)
  - `npm --workspace @microsoft/kiota-serialization-text run test:node` (5/5 passed)
- Monorepo build and formatting:
  - `npm run build` (all 9 packages built cleanly)
  - `npm run prettier:check` (all matched files formatted)
  - `npm run lint` (0 errors, 0 warnings)
- Test suite:
  - `npm run test:node` (all 9 package test suites passed)


packages/serialization/json/src/jsonParseNode.ts
```diff
@@ -55,16 +55,32 @@ export class JsonParseNode implements ParseNode {
...
+			const typeOfValue = typeof value;
+			if (typeOfValue === "boolean" || typeOfValue === "string" || typeOfValue === "number") {
+				return value as unknown as T;
+			} else if (value instanceof Date) {
+				return this.getDateValueFromRaw(value) as unknown as T;
+			} else if (value instanceof DateOnly) {
+				return this.getDateOnlyValueFromRaw(value) as unknown as T;
+			} else if (value instanceof TimeOnly) {
+				return this.getTimeOnlyValueFromRaw(value) as unknown as T;
+			} else if (value instanceof Duration) {
```
baywet, line 78: instanceof has a lot of pitfalls, and performance implications

packages/serialization/text/src/textParseNode.ts
```diff
@@ -75,6 +75,7 @@ export class TextParseNode implements ParseNode {
 				case "Duration":
 					return node.getDurationValue() as unknown as T;
 				case "string":
+				case undefined:
```
baywet, line 78: we're going to lose fidelity for any string based representation here, and create runtime issues with the auto-boxing

baywet (changes requested):
Thanks for the contribution!

No, making the parameter optional is not the correct solution here.
TypeScript is not a stable language, so we're allowed to make breaking changes for the time being.
If the previously generated client is misaligned, it should be re-generated with the latest version of kiota.
If kiota is "omitting" to generate the type parameter, that should be fixed in the generator itself.

rksharma-owg:
Thanks for clarifying the compatibility policy. I’ll keep the parse-node parameter required and trace the missing primitive type argument through Kiota’s generator instead of adding runtime fallback behavior in the TypeScript runtime.

packages/serialization/json/src/jsonParseNode.ts
rksharma-owg: Understood. I’ll remove runtime-type dispatch from the solution; the generated call should provide the primitive type so this parser can stay on its existing explicit dispatch path.

packages/serialization/text/src/textParseNode.ts
rksharma-owg: Agreed—the undefined/string fallback would erase the intended primitive distinction and interfere with auto-boxing. I’ll address the omitted argument at generation time instead.

rksharma-owg:
Follow-up: I traced this through the generator. microsoft/kiota#7800 was merged on June 18 and adds the required primitive type token to generated TypeScript collection deserializers. The current `@microsoft/msgraph-sdk@1.0.0-preview.89` now emits `getCollectionOfPrimitiveValues("string")` for `subscribedSku.subscriptionIds`, and I verified that it deserializes correctly with `@microsoft/kiota-serialization-json@1.0.0-preview.107`. No additional runtime fallback is needed; affected consumers should update or regenerate their client. I’ll leave this closed PR unchanged.

PR #2137, opened 2026-09-14
feat(http): add fetch request option for credentials and native fetch configuration

### LINKED ISSUE #1817: Use secure cookie with the native fetch credentials: 'include'
Hi,
I try Kiota to generate a client usable by my frontend.
However my frontend use secure cookie to connect to my HTTP REST API.
Fetch already have a [native option](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) to include secure cookie in the request but Kiota doesn't seems to give a way to configure fetch options out of the box when creating the client.
As it took me some time to figure out the right Kiota configuration, I wanted to document this case here.

```ts
import { AnonymousAuthenticationProvider } from '@microsoft/kiota-abstractions';
import { DefaultRequestAdapter } from '@microsoft/kiota-bundle';
import { KiotaClientFactory } from '@microsoft/kiota-http-fetchlibrary';

import { createClient } from './generated/client.js';

// API requires no authentication, so use the anonymous
// authentication provider
const authProvider = new AnonymousAuthenticationProvider();

// Define a custom fetch function to set the 'credentials' option.
// This function wraps the global fetch, but adds our required option.
const customFetch = (url: Parameters<typeof fetch>[0], init?: RequestInit) => fetch(url, {
  ...init,
  credentials: 'include', // 'include' tells the browser to send cookies
});

const httpClient = KiotaClientFactory.create(customFetch);
const adapter = new DefaultRequestAdapter(authProvider, undefined, undefined, httpClient);
export const createHomaioClient = (baseUrl: string) => {
  // Remove trailing slash
  adapter.baseUrl = baseUrl.replace(/\/$/, '');
  return createClient(adapter);
};
```

It may be valuable to add this case to the documentation.

rksharma-owg:
Resolves #1817

### Summary

Adds `FetchRequestOption` to `@microsoft/kiota-http-fetchlibrary` to allow configuring native fetch `RequestInit` options (such as `credentials: 'include'`, `mode`, `cache`, `keepalive`, etc.) both per-request and as default adapter options. Also documents configuring credentials for secure cookies using either `FetchRequestOption` or a custom fetch function with `KiotaClientFactory.create(customFetch)`.

### Changes

1. **`FetchRequestOption`**:
   - Implements `RequestOption` with key `FetchRequestOptionKey`.
   - Allows configuring `credentials` (`'include' | 'same-origin' | 'omit'`), `mode`, `cache`, `integrity`, `keepalive`, `redirect`, `referrer`, `referrerPolicy`, and arbitrary custom `init` properties.

2. **`FetchRequestAdapter`**:
   - Accepts `defaultFetchOptions?: FetchRequestOption` in the constructor and exposes it as a public property.
   - Merges `defaultFetchOptions` and request-level `FetchRequestOption` into the `RequestInit` in `getRequestFromRequestInformation` / `convertToNativeRequest`.

3. **Exports & Documentation**:
   - Exported in root `src/index.ts` and `src/browser/index.ts`.
   - Added documentation and code examples to `packages/http/fetch/README.md`.

4. **Testing**:
   - Added unit tests in `packages/http/fetch/test/common/middleware/fetchRequestOptionTest.ts` verifying default initialization, custom option passing, request-level options, default adapter options, and request overrides.

### Verification
- `npm run prettier:check` passed
- `npm run lint` passed
- `npm run test:node` passed across all 9 workspaces
- `npm run build` passed across all workspaces


packages/http/fetch/test/common/middleware/fetchRequestOptionTest.ts
```diff
@@ -0,0 +1,113 @@
...
+		assert.equal(option.referrerPolicy, "no-referrer");
+		assert.deepEqual(option.init, { customOption: "customValue" });
+	});
+
+	it("Should apply FetchRequestOption to native request in convertToNativeRequest", async () => {
+		const adapter = new FetchRequestAdapter(new AnonymousAuthenticationProvider());
+		const requestInfo = new RequestInformation();
+		requestInfo.URL = "https://example.com/api";
+		requestInfo.httpMethod = HttpMethod.POST;
+		requestInfo.content = "test body";
```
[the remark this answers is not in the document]
rksharma-owg: Updated the test to set `requestInfo.content` to an `ArrayBuffer` using `TextEncoder`, matching the type definition.

packages/http/fetch/README.md
```diff
@@ -12,6 +12,51 @@ Read more about Kiota [here](https://github.com/microsoft/kiota/blob/main/README
...
+## Configuring Native Fetch Options (e.g., Cookies / Credentials)
+
+If your API requires sending cookies or credentials (such as `credentials: 'include'`), or custom options like `mode`, `cache`, or `keepalive`, you can configure them in two ways:
+
+### 1. Using `FetchRequestOption`
+
+Attach `FetchRequestOption` either globally on `FetchRequestAdapter` or per request:
+
+```typescript
+import { FetchRequestOption } from '@microsoft/kiota-http-fetchlibrary';
```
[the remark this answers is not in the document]
rksharma-owg: Changed the check to `fetchOption.referrer !== undefined` (and similarly for `referrerPolicy`) so that empty strings or other valid values are preserved and can override default options. Added a unit test to verify.

packages/http/fetch/src/fetchRequestAdapter.ts
```diff
@@ -538,12 +540,50 @@ export class FetchRequestAdapter implements RequestAdapter {
...
+		if (fetchOption.integrity) {
+			req.integrity = fetchOption.integrity;
+		}
+		if (fetchOption.keepalive !== undefined) {
+			req.keepalive = fetchOption.keepalive;
+		}
+		if (fetchOption.redirect) {
+			req.redirect = fetchOption.redirect;
+		}
+		if (fetchOption.referrer) {
```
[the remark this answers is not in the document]
rksharma-owg: Added `FetchRequestAdapter` to the import statement in the README code example.

packages/http/fetch/src/middlewares/options/fetchRequestOption.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import { type RequestOption } from "@microsoft/kiota-abstractions";
+
+export const FetchRequestOptionKey = "FetchRequestOptionKey";
+
+export type FetchCredentials = "omit" | "same-origin" | "include";
+export type FetchMode = "cors" | "navigate" | "no-cors" | "same-origin";
+export type FetchCache = "default" | "force-cache" | "no-cache" | "no-store" | "only-if-cached" | "reload";
+export type FetchRedirect = "error" | "follow" | "manual";
+export type FetchReferrerPolicy = "" | "no-referrer" | "no-referrer-when-downgrade" | "origin" | "origin-when-cross-origin" | "same-origin" | "strict-origin" | "strict-origin-when-cross-origin" | "unsafe-url";
```
baywet: could we get this from the node/dom type definitions instead of redefining here?

packages/http/fetch/src/fetchRequestAdapter.ts
```diff
@@ -44,6 +46,7 @@ export class FetchRequestAdapter implements RequestAdapter {
 		private readonly httpClient: HttpClient = new HttpClient(),
 		observabilityOptions: ObservabilityOptions = new ObservabilityOptionsImpl(),
 		private backingStoreFactory = new InMemoryBackingStoreFactory(),
+		public defaultFetchOptions?: FetchRequestOption,
```
baywet: having both this and a middleware handler kind of duplicates the API experience. What lead you to that design choice?

rksharma-owg: The intent was to support two scopes: adapter-wide defaults for every request, and a `FetchRequestOption` attached to `RequestInformation` for per-request overrides. The implementation currently applies both in `getRequestFromRequestInformation`; it does not add a separate middleware handler.

That said, the adapter-level property/constructor argument is not necessary for the original cookie use case: `KiotaClientFactory.create(customFetch)` already handles global defaults. A middleware could also own default/per-request merging without extending the adapter constructor.

Would you prefer that I move the behavior into a middleware with constructor defaults and per-request `FetchRequestOption` overrides, or narrow this PR to documenting the existing custom-fetch solution from #1817? I will hold the API restructuring until that direction is clear.

packages/http/fetch/src/middlewares/options/fetchRequestOption.ts
rksharma-owg: Yes. These should come from `RequestInit` rather than duplicate its unions—for example, `NonNullable<RequestInit["credentials"]>` for the existing `FetchCredentials` alias, and the corresponding indexed types for mode, cache, redirect, and referrerPolicy. The project already includes DOM types. I will make that adjustment with the agreed API shape from the other thread, so the next update is one coherent correction.

packages/http/fetch/src/fetchRequestAdapter.ts
baywet: yes, I believe that providing consumers multiple ways to do the same thing is only going to create confusion and increase maintenance cost.
Since they already have a way to specify fetch options, through the custom fetch approach, let's make sure this path is obvious to discover for those who need it. And re-evaluate based on user feedback if that approach poses important limitations.

packages/http/fetch/src/middlewares/options/fetchRequestOption.ts
rksharma-owg: The narrowed update removes `FetchRequestOption` and its type aliases entirely, so there are no duplicated DOM unions left to maintain.

packages/http/fetch/src/fetchRequestAdapter.ts
rksharma-owg: Narrowed the PR to the existing custom-fetch path. The new adapter option API and tests were removed; the remaining one-file change documents `KiotaClientFactory.create(customFetch)` with `credentials: "include"`. The exact head `54937f2` passed the full fork preflight matrix.

baywet (dismissed):

packages/http/fetch/README.md
```diff
@@ -12,6 +12,25 @@ Read more about Kiota [here](https://github.com/microsoft/kiota/blob/main/README
 
 1. `npm i @microsoft/kiota-http-fetchlibrary`.
```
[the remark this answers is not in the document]
baywet: @rksharma-owg Would you mind applying that change please?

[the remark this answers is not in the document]
rksharma-owg: Updated the install command to include both packages imported by the example in 2ce06c9. The exact commit passed the fork preflight across Node 20/22/24, including build, lint, integrated build, tests, and package coverage. Upstream checks are running now.

baywet (approved):

PR #2149, opened 2026-09-22
fix: correct host parsing in AllowedHostsValidator

Treicysg:
This PR refactors AllowedHostsValidator.isUrlHostValid  to use the standard URL  parser instead of manual string splitting when extracting the host from a request URL. This makes host extraction more robust and consistent with how the other Kiota language SDKs handle this check.


packages/abstractions/src/authentication/allowedHostsValidator.ts
```diff
@@ -42,31 +42,25 @@ export class AllowedHostsValidator {
 	public isUrlHostValid(url: string): boolean {
 		if (!url) return false;
 		if (this.allowedHosts.size === 0) return true;
-		const schemeAndRest = url.split("://");
-		if (schemeAndRest.length >= 2) {
-			const rest = schemeAndRest[1];
-			if (rest) {
-				return this.isHostAndPathValid(rest);
+		const candidate = url.includes("://") ? url : !url.startsWith("http") ? `http://${url}` : undefined;
```
baywet: why are you defaulting to http? and not https? or better yet, if we're in browser, take the browser current protocol to implement properly protocol relative uris

baywet (changes requested):

packages/abstractions/src/authentication/allowedHostsValidator.ts
Treicysg: Good catch, updated to use  window.location.protocol  when running in a browser, falling back to https otherwise.

baywet (approved):

packages/abstractions/src/authentication/allowedHostsValidator.ts
Treicysg: qq, should we go further and drop scheme-less/protocol-relative URL support entirely, returning false ?. I think that is what kiota-python does

